PR #58, opened 2016-02-18
Fix #40: Add custom metric for document.compatMode

### LINKED ISSUE #40: Collect quirks mode
It would be useful to store which rendering mode was used for the top-level document. There is a `doctype` field but it is not obvious how to map that to quirkyness and it will not be quite correct.

document.compatMode === 'BackCompat' // quirks mode
document.compatMode === 'CSS1Compat' // standards mode or almost standards mode

zcorpan:


stevesouders:
This change will add the data to the WPT results. Is that all you wanted? 

It's a lot more work to add a new stat all the way through the pipeline. That involves:

1. changing the schema to store the raw data
2. changing the schema to store aggregate results (we don't query the raw results in real time)
3. updating the schema in the dev & production databases
4. writing code to store the raw results
5. writing code to store the aggregate results
6. modifying the UI to display the aggregate results
7. coordinating the push of this new data to BigQuery

I'd be happy to accept a pull request for 1, 2, 4, 5, & 6. I can take care of 3 & 7.

pmeenan:
This would put it in the HAR files which could be queried though. Probably a good first step for proving usefulness before fully plumbing it. 






On Thu, Feb 18, 2016 at 5:29 PM -0800, "Steve Souders" <notifications@github.com> wrote:












This change will add the data to the WPT results. Is that all you wanted? 



It's a lot more work to add a new stat all the way through the pipeline. That involves:


changing the schema to store the raw data
changing the schema to store aggregate results (we don't query the raw results in real time)
updating the schema in the dev & production databases
writing code to store the raw results
writing code to store the aggregate results
modifying the UI to display the aggregate results
coordinating the push of this new data to BigQuery




I'd be happy to accept a pull request for 1, 2, 4, 5, & 6. I can take care of 3 & 7.

zcorpan:
I only really care about it being present in BigQuery, personally.

zcorpan:
Should it be stored as a boolean instead of a string?

pmeenan:
String is probably best since there are multiple modes and this provides
the most detail.  To be clear, it won't be in the BigQuery tables, it will
be queryable through the new work Ilya is doing to have the full HAR files
in json in the results in BigQuery (not sure if he has documented how to
query that yet).

Getting it into the tables would require steps 1-3 in Steve's list but I
don't think we want to do that right away.  Having it in the HAR's and
being able to query it through the HAR query framework Ilya is putting
together first to work through use cases will give us an idea if that is
good enough or if it is something we wand in the database tables as well.

In the HAR files it will look something like
this: "_compat_mode":"CSS1Compat"

You can click the "Export HAR" link on the WPT test here to see the full
HAR:
http://www.webpagetest.org/result/160219_6W_8db94a17b321ead7c63039b12a71914b/

zcorpan:
> String is probably best since there are multiple modes and this provides
the most detail.  

But it can only be two different values. "BackCompat" or "CSS1Compat".

> To be clear, it won't be in the BigQuery tables, it will
be queryable through the new work Ilya is doing to have the full HAR files
in json in the results in BigQuery (not sure if he has documented how to
query that yet).

The new JSON stuff is what I want. :-)

pmeenan:
I'm fine either way.  If you would prefer a bool then I'd probably prefer
it be called quirks_mode instead.  I kind of like having it as a string for
future-proofing though in case more get added at some point.

zcorpan:
I think it's very, very, very unlikely that compatMode ever returns a new string. We already have 4 modes (quirks, limited quirks, no quirks, XML) and this API only returns two values. Some versions of IE have like 73 browser modes and no change to this API.

OK I'll change it to quirks_mode...

zcorpan:
Done.

pmeenan:
Steve, do you have a script we need to run on the server to have this included in the next crawl?

souders:
I just need to push it by March 1.

-Steve

souders:
Done pushing. This result should be in the results for the March 1 crawl.

-Steve

zcorpan:
Thank you!

PR #60, opened 2016-03-27
Api

alvarodlg:


rviscomi:
In https://github.com/HTTPArchive/httparchive/issues/81 we're planning a re-architecture of HA and the underlying data/metrics are liable to change. We should revisit the idea of an API when that settles. Although it's not very web-accessible, we always have the queryable raw data on BigQuery.

igrigorik:
+1. We have some documentation on various ways to access the data here: https://discuss.httparchive.org/t/quickstart-guide-to-exploring-the-http-archive/682

Closing this. Let's revisit once we land the new design+metrics.

Issue #77, opened 2017-03-22
Detect third party JS libraries with custom metric


rviscomi:
Using custom metric scripts, detect the presence of third party libraries and their version if available.

For example, a page may have "jquery@3.2.0,modernizr,backbone@1.3.3,...".

rviscomi:
Previous discussion: https://discuss.httparchive.org/t/tracking-javascript-library-versions-in-http-archive/55/24

tlauinger:
I'm a co-author of the paper referenced in #78. Our library detection code is a fork of the [library detection code](https://github.com/johnmichel/Library-Detector-for-Chrome/blob/master/library/libraries.js) used by the [Library Detector for Chrome](https://github.com/johnmichel/Library-Detector-for-Chrome) extension. We did a few modifications to the code more than a year ago, so our fork isn't really up to date any more. I compared our code to the [current version](https://github.com/johnmichel/Library-Detector-for-Chrome/blob/c8b672714343e6dd59f5c5fa229b80c1b5bc0a41/library/libraries.js) today and found the following differences:

_Libraries detected by the Chrome extension but not by our code:_ GWT, Ink, Vaadin, Zurb, Polymer, Highcharts, InfoVis, Blackbird, CreateJS, Google Maps, Spry, Qooxdoo, Ext JS, base2, closure, Processing.js, Mapbox, Sammy, Rico, MochiKit, gRaphaël (fix `gRapha&euml;l`), Glow, FuseJS, Tween.js, SproutCore, Zepto.js, PhiloGL, LABjs, Head JS, ControlJS, RightJS, Pusher, Swiffy, Move, AmplifyJS, Popcorn.js, Spine, Visibility.js, IfVisible.js, DC.js, Vue, Two, Brewser, Material Design Lite, Kendo UI, Matter.js, Riot, Sea.js, ScrollMagic

_Libraries not currently detected by the Chrome extension (added by us):_ swfobject, flexslider, moment-timezone (plug-in), json3, pep, spf, numeral

_Equivalent detection code in both:_ Lo-Dash, Underscore (test checks that no Lo-Dash detection in window), yepnope, jQuery Tools, D3, Ember.js, Greensock JS, Isotope

_Detection code better in the Chrome extension (better error handling etc.):_ Bootstrap, three.js (better handling of two cases), CamanJS, WebFont Loader (set version to `null`)

[...]

PR #78, opened 2017-03-24
custom metric for 3rd party libs

rviscomi:
Proof of concept of #77 

First library is jQuery, can add more incrementally.

Inspired by http://www.ccs.neu.edu/home/arshad/publications/ndss2017jslibs.pdf
/cc toby@ccs.neu.edu

### DIFF
--- a/custom_metrics/third-parties.js
+++ b/custom_metrics/third-parties.js
@@ -1,0 +1,34 @@
+/**
+ * Detect the presence of third party libraries.
+ *
+ * Look for common library aliases on the global scope.
+ * If available, detect libraries' versions.
+ *
+ * Outputs JSON-serialized list of library-version pairs.
+ *     e.g.: `["a@1.0","b","c@2.2"]`
+ */
+var thirdParties = [];
+var name, version;
+
+function addThirdParty(name, version) {
+	if (!name) {
+		return;
+	}
+
+	if (version) {
+		name += '@' + version;
+	}
+
+	thirdParties.push(name);
+}
+
+// jQuery
+// https://github.com/jquery/jquery/blob/c1c549793a8772107e128d21f8a8f0c3fdf0f027/src/core.js#L53
+var jQuery = window.jQuery || window.$ || window.$jq || window.$j;
+name = jQuery && jQuery.fn && 'jQuery';
+version = name && jQuery.fn.jquery
+addThirdParty(name, version);
+
+// TODO(rviscomi): Add more libraries...
+
+return JSON.stringify(thirdParties);


custom_metrics/third-parties.js
```diff
@@ -0,0 +1,34 @@
+[third-parties]
```
igrigorik: Is this intentional? Looks like a syntax error :)

custom_metrics/third-parties.js
```diff
@@ -0,0 +1,34 @@
...
+		name += '@' + version;
+	}
+
+	thirdParties.push(name);
+}
+
+// jQuery
+var jQuery = window.jQuery || window.$ || window.$jq || window.$j;
+name = jQuery && jQuery.fn && 'jQuery';
+version = name && jQuery.fn.jquery
```
igrigorik: wishlist: link to the library source where this is defined, so we have a reference / can track changes more easily.

pmeenan: That's how the block definition works for WebPageTest when submitting through the UI or API but not for permanent ones installed on the server.  Those should just be scripts with one script per metric.

i.e. yes, it should be removed.

custom_metrics/third-parties.js
pmeenan: That's how the block definition works for WebPageTest when submitting through the UI or API but not for permanent ones installed on the server. Those should just be scripts with one script per metric.

i.e. yes, it should be removed.

rviscomi: Oops. Yeah I had it in there for testing on the UI. Removed.

custom_metrics/third-parties.js
rviscomi: Done

igrigorik:
@pmeenan @rviscomi  how do we ensure that new custom metrics are automatically picked up in subsequent runs?

rviscomi:
Is there an existing process for updating the private WPT test server? If not we can either manually update the custom_metric dir or formalize some kind of wholesale WPT update process.

pmeenan:
@stevesouders had a process for syncing the custom metrics over (not sure how automated it was).

custom_metrics/third-parties.js
```diff
@@ -0,0 +1,34 @@
...
+var thirdParties = [];
+var name, version;
+
+function addThirdParty(name, version) {
+	if (!name) {
+		return;
+	}
+
+	if (version) {
+		name += '@' + version;
```
tlauinger: I would recommend finding a different way to serialise the library name/version pairs. In the wild, there are lots of "unexpected" version strings that will cause issues later. Typically, one would expect something like `1.2.3`, but it is also very common to have `1.3.0-rc.4`, and then (less common but still frequent in large crawls) many constructs such as `1.4.11-local+sha.2c9066e`, `@EDGE` or `@VERSION` or `@@VERSION`, `0.0.0dev (22487)`, `1.9.7build20160503-0010,1.9.7-20130926-IBM (2dfbbaf)`, `0.0.0${project.version} (23843)`, `2.2.3 -ajax/load,-ajax/script,-ajax/jsonp,-deprecated,-effects,-effects/animatedSelector,-effects/Tween,-event/alias,-wrap,-core/ready,-css/hiddenVisibleSelectors` (looks like a custom build of jQuery), `$Rev: 249 $`, and some versions of jQuery Tools return not a string, but a dict/object of all installed plug-ins, which can look like this: `{'tabs': '1.0.1', 'scrollable': '1.0.5', 'expose': '1.0.3', 'tooltip': '1.0.2', 'overlay': '1.0.4'}`.

In our paper, we ignored these special versions because we usually cannot find metadata (such as release dates) for them, and it's very complicated (or impossible) to compare if one of these special versions is older or newer than another. For a general-purpose web crawl, I think it makes sense to keep all versions, but if these special version strings/objects aren't properly "escaped", splitting the string into name/version in a DB query will get quite complicated.

tlauinger: I think it would also make DB queries easier if the version part was always appended (even if unknown/null), so that there's no need to distinguish in the query between the two cases "name" and "name@version", because it would either be "name@1.2.3" or "name@unknown".

tlauinger:
I added a few comments to the PR, but noticed that it has already been merged

igrigorik:
> for a general-purpose web crawl, I think it makes sense to keep all versions, but if these special version strings/objects aren't properly "escaped", splitting the string into name/version in a DB query will get quite complicated.

+1. We should store as much as we can at crawl time; we can extract what we need for analysis later.

igrigorik:
> @stevesouders had a process for syncing the custom metrics over (not sure how automated it was).

AFAIK, we talked about automating it, but never did.. What are the spots where we'd need to push the updates? Main server + all the agents? Rick, let's split this into a separate bug?

pmeenan:
Just the server under /var/www/webpagetest/settings/custom_metrics (or something like that).  Server pushes them down to the agents along with the test data

stevesouders:
My technique is:
cd /var/www/httparchive.dev/
make pushmetrics

rviscomi:
Thanks @stevesouders ! Is httparchive.dev the local git repo? If so then the process as I understand would be:

1. ssh into HA
2. `cd /var/www/httparchive.dev/`
3. `git pull`
4. `make [pushmetrics](https://github.com/HTTPArchive/httparchive/blob/814502e7ac84f61c46b96275bce1b4cf092150af/Makefile#L99)`

This way the latest changes to the public repo are copied to the local WPT dir. Sound right?

@igrigorik WDYT of also automating this with a cron job? Either way, this process should be immortalized in the docs.

igrigorik:
@rviscomi sgtm, there is already a cron job that kicks off the crawl, we should a step to it..

rviscomi:
> I think it would also make DB queries easier if the version part was always appended (even if unknown/null), so that there's no need to distinguish in the query between the two cases "name" and "name@version", because it would either be "name@1.2.3" or "name@unknown".

See https://github.com/HTTPArchive/httparchive/pull/80. Empty versions would be `name@null`.

PR #80, opened 2017-03-28
always append a version, even if null

rviscomi:
See https://github.com/HTTPArchive/httparchive/pull/78#discussion_r108254522

Example WPT: https://www.webpagetest.org/result/170328_CA_1G4M/1/details/#waterfall_view_step1
Example source code: http://codepen.io/anon/pen/jBemrB

Progress on https://github.com/HTTPArchive/httparchive/issues/77

### DIFF
--- a/custom_metrics/third-parties.js
+++ b/custom_metrics/third-parties.js
@@ -15,9 +15,7 @@ function addThirdParty(name, version) {
 		return;
 	}
 
-	if (version) {
-		name += '@' + version;
-	}
+	name += '@' + (version || 'null');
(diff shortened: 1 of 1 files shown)


igrigorik:
@rviscomi nitpick: "genthirdpartymetrics" is a very generic concept / doesn't describe what it does. Should we rename this stuff to ~detect-libraries or some such?

rviscomi:
Forgot that pushing commits while a PR is open updates the PR 😜  Moving to separate branch/PR.

PR #85, opened 2017-04-03
Generate third party detection

rviscomi:
`npm run custom-metrics` generates third party detection code based on https://github.com/johnmichel/Library-Detector-for-Chrome/blob/master/library/libraries.js.

Progress on #77


pmeenan:
That's a LOT of code to be sending around with every test.  At a minimum can you strip out the icon and url references for each library?  It would probably be a good idea to minify it as well (just spaces and comments).

As a sanity check, did you make sure it works on WebPageTest as a custom metric?

It also looks like some of the versions have spaces in the name - just want to make sure it is expected and you'r format for transmitting/querying will be ok with that

rviscomi:
> That's a LOT of code to be sending around with every test. At a minimum can you strip out the icon and url references for each library? It would probably be a good idea to minify it as well (just spaces and comments).

You're right, I'll look into that.

> As a sanity check, did you make sure it works on WebPageTest as a custom metric?

LGTM https://www.webpagetest.org/result/170403_ZV_19Q8/1/details/#waterfall_view_step1

> It also looks like some of the versions have spaces in the name - just want to make sure it is expected and you'r format for transmitting/querying will be ok with that

Spaces should be ok. Maybe we do need to escape `@` so it can be used as a delimiter.

rviscomi:
Stripping out icon/url properties and comments got the file down from 33.4 KB to 26.3 KB. Reduced further to 17.4 KB by collapsing whitespace. I'll run another test to make sure these don't break.

bin/library-detection.js
```diff
@@ -0,0 +1,105 @@
...
+const LIBRARY_DETECTOR_URL = 'https://raw.githubusercontent.com/johnmichel/Library-Detector-for-Chrome/master/';
+const LIBRARY_DETECTOR_MAIN_URL = `${LIBRARY_DETECTOR_URL}/library/libraries.js`;
+const LIBRARY_DETECTOR_MANIFEST_URL = `${LIBRARY_DETECTOR_URL}/manifest.json`;
+const LIBRARY_DETECTOR_VAR_PATTERN = /([a-f0-9]*_LibraryDetectorTests)/;
+const TMP_FILE = 'tmp/libraryDetector.js';
+const VERSION_METRIC_FILE = 'custom_metrics/lib-detector-version.js';
+const OUTPUT_FILE = 'custom_metrics/third-parties.js';
+const OUTPUT_HEAD = `/**
+ * IMPORTANT: Do not modify this file directly! It is generated by
+ *   bin/library-detection.js
```
igrigorik: nitpick 2: bin/library-detector to stay consistent? ;-)

bin/library-detection.js
```diff
@@ -0,0 +1,105 @@
...
+var thirdParties = [];
+
+function addThirdParty(name, version) {
+	if (!name) {
+		return;
+	}
+
+	name = name.replace('@', '\\@');
+	version = version && version.replace('@', '\\@');
+	name += '@' + (version || 'null');
```
igrigorik: Hmm, how BigQuery friendly is this? We should test how easy/hard it will be to work with these pairs in BQ. I imagine we'll want to be doing [split()](https://cloud.google.com/bigquery/docs/reference/legacy-sql#split)'s on these strings at some point.. Will the escaped chars cause problems?

FWIW, BigQuery can handle nested records. Are we limited by WPT's custom metric format? If not, perhaps we can output an array of well formed (serialized) objects?

pmeenan:
WPT just takes the string blob returned and passes it through.  It can be anything you want.  It would also be trivially easy to make it understand that a response is JSON and decode it so that decoded JSON is available directly in the HAR (instead of an escaped JSON string)

bin/library-detection.js
rviscomi: I thought the split might be able to take a pattern like `[^\]@`. Also considered replacing `@` with `<at>` or similar.

I agree that it could be made more BQ friendly. For example, an array of objects similar to `{name: 'jQuery', version: '1.2.3'}`.

pmeenan:
Is the code planning to detect multiple installs of the same library?  If not, using the library name as the dictionary keys may be cleaner: ```{"jQuery": "1.2.3"}```

ebidel:
One question I have is how well the detections will work in practice? For example, many people create bundles using browserify or webpack that keep libraries off `window`. e.g. `window.Polymer.version` will fail.

rviscomi:
> Is the code planning to detect multiple installs of the same library? If not, using the library name as the dictionary keys may be cleaner: `{"jQuery": "1.2.3"}`

No immediate plans, but it's plausible. I like the conciseness of your suggestion. Is it as easy to query as `foo.name` and `foo.version`?

> One question I have is how well the detections will work in practice? For example, many people create bundles using browserify or webpack that keep libraries off window. e.g. `window.Polymer.version will fail.`

Yeah that's definitely a limitation. See @tlauinger's doc for more info https://github.com/HTTPArchive/httparchive/blob/master/docs/js-library-detection.md#limitations-of-the-detection-methodology. It'd be interesting to take a sample from a complete run and manually inspect the sites to get a sense of detection accuracy.

bin/library-detection.js
rviscomi: Done

igrigorik:
> It would also be trivially easy to make it understand that a response is JSON and decode it so that decoded JSON is available directly in the HAR (instead of an escaped JSON string)

That sounds nice. How would we indicate that we're emitting a JSON object in the custom metrics pipeline?

>> Is the code planning to detect multiple installs of the same library? If not, using the library name as the dictionary keys may be cleaner: {"jQuery": "1.2.3"}
> No immediate plans, but it's plausible. I like the conciseness of your suggestion. Is it as easy to query as foo.name and foo.version?

The downside is lack of extensibility? The nice thing about the bag approach is that we can easily stuff more keys into it down the road, if we ever need to. On the other hand, querying is definitely harder.. ¯\_(ツ)_/¯

tlauinger:
Currently, the code can detect only a single instance of a library per page. Extending it to detect libraries in frames is relatively easy, and would already result in some kind of duplicates (same page, different frame). There are also ways to detect "true" duplicates (same page, same frame), even though the easier ways of achieving this wouldn't need to run during the live crawl. Another possible extension point is to link library detections to resources/URLs. At the moment, a library detection doesn't reveal where the library was loaded from, or how it was included, only that it's present in a JS environment.

It's difficult to anticipate if or how any of these extensions will ever get implemented. Each of these extensions would require a slightly different representation. To support duplicates, it's enough to have a list of version strings for each library key, but for resource linking, one would also need to integrate URLs into the data structure. Major changes in detection functionality would likely require the data structure to be modified anyway.

igrigorik:
> Another possible extension point is to link library detections to resources/URLs. At the moment, a library detection doesn't reveal where the library was loaded from, or how it was included, only that it's present in a JS environment.

+1, that seems like a great future feature, and a good argument for sticking with bag of values approach.

rviscomi:
Updated to use array of objects.

```js
function addThirdParty(name, version='null') {
	if (!name) {
		return;
	}

	thirdParties.push({name, version});
}
```

WPT LGTM: https://www.webpagetest.org/result/170405_58_1595/1/details/#waterfall_view_step1

bin/library-detector.js
```diff
@@ -0,0 +1,102 @@
...
+ *   bin/library-detector.js
+ *
+ * Detects the presence of third party libraries.
+ *
+ * Look for common library aliases on the global scope.
+ * If available, detect libraries' versions.
+ * Built on https://github.com/johnmichel/Library-Detector-for-Chrome.
+ *
+ * Outputs JSON-serialized list of library-version pairs.
+ *     e.g.: \`[{name: "a", version: "1.0"},{name: "b", version: "null"}]\`
```
igrigorik: Is the "null" meant to be a string, or an actual null JSON value?

rviscomi: JSON/BQ can handle `null` so changed to use that instead.

rviscomi:
Yeah I'll squash+merge

igrigorik:
Woot! Can't wait to see the data from the next HA run... 👍

PR #94, opened 2017-04-17
add LICENSE

rviscomi:


rviscomi:
silly me. we have one, but named COPYING. i'll just _copy_ it.

PR #97, opened 2017-04-18
Generate CMS custom metric

rviscomi:
- use AliasIO/Wappalyzer tests
- timestamp of generation for versioning
    - repo lacks a version string
- add license info to both CMS and lib detection

WPT: https://www.webpagetest.org/custom_metrics.php?test=170418_5K_1730&run=1&cached=0
script: https://gist.github.com/rviscomi/f6c71a7949d339f1a13dfb0fc46f2fbb

Stripped Wappalyzer detection code to only required data. `cms.js` is 14.4 KB. (`third-parties.js` is 18.6 KB)

Progress on https://github.com/HTTPArchive/httparchive/issues/90


igrigorik:
Per discussion earlier today.. 
- it'd be nice to support all of the wappalyzer tests
- it'd be nice to be able to run same logic on past runs

An alternative to custom metric approach is to integrate this detection into our dataflow pipeline, either as an extra step for the import pipeline or as a separate pipeline. If we were to do that.. would we still want the custom metric? The other benefit of Dataflow approach is that we can access headers, which some of the tests require.

rviscomi:
I'll close this PR for now as I investigate the dataflow approach. The code will live on in [my branch](https://github.com/rviscomi/httparchive/commit/57a299422e5b9950042371e095522279dc5185b4).

Issue #79, opened 2017-03-28
Add documentation


rviscomi:
Document the development and maintenance flow. Use prosaic text, diagrams, and inline comments as necessary.

Open questions:
- In what format do we write the docs: Markdown? HTML? Plain text?
- Where do the docs live: /docs dir? Google Docs? Adjacent to code?
- Do we surface the docs on the web interface? How?
- What are the high priority things to document?
- What is the process for updating docs?

rviscomi:
> In what format do we write the docs: Markdown? HTML? Plain text?
> Where do the docs live: /docs dir? Google Docs? Adjacent to code?
> Do we surface the docs on the web interface? How?

These three are tightly connected.

One idea is to create markdown files in a /docs subdirectory and create a build rule for generating a web-friendly docs site in HTML.

pmeenan:
Also need to separate out things that are generic for any httparchive install (if people are still running them) vs anything that is specific to the httparchive.org instance

igrigorik:
FWIW, I'm not aware of anyone running this on their own. I don't think we should discourage that (it is open source), but I don't think that should be our goal either. For example, we'll be moving a lot of our processing away from MySQL to BigQuery..

> In what format do we write the docs: Markdown? HTML? Plain text?
> Where do the docs live: /docs dir? Google Docs? Adjacent to code?
> Do we surface the docs on the web interface? How?

My vote is for markdown, as it's easy to author and maintain. To start, I'd propose we create /docs and put everything there. Longer term, I'd love to see these docs visible on httparchive.org/about/{something}.. we can move MD files once we have the new design.

rviscomi:
Working on a draft of a doc on the HA data pipeline. Will check it in as .md when ready, but would appreciate comments/edits to help fill in the blanks or make corrections.

https://docs.google.com/document/d/1z18KGAJZ0i4WCYF06nlvuT5_284Idt1_JIYMb3JLSXg/edit?usp=sharing

igrigorik:
Nice! Great start, looking forward to fleshing it out.

rviscomi:
@paulcalvano is going to work on adding an "Getting Started with BigQuery" doc. (Can't assign to him for some reason 😭).

igrigorik:
I think Paul needs to be a member of the org..

rviscomi:
He is! :sob: :sob:

paulcalvano:
[...]

PR #110, opened 2017-05-22
Archive docs

rviscomi:
Progress on #79


docs/archive.md
```diff
@@ -0,0 +1,26 @@
+# Data Archiving
+
+When the crawls are complete, the data is archived to several places:
+
+- Internet Archive (sql)
```
igrigorik: should probably clarify that it's "MySQL dumps"?

docs/archive.md
```diff
@@ -0,0 +1,26 @@
...
+- HTTP Archive (sql, csv)
```
igrigorik: This one seems circular? :)

docs/archive.md
```diff
@@ -0,0 +1,26 @@
...
+- Google Storage (har)
+- BigQuery (har)
+
```
igrigorik: FWIW, I don't think we need to document every surface here, as much as.. points folks towards the canonical place they should prefer and can rely on. For that, I propose we use gs://.

docs/bigquery.md
```diff
@@ -1,5 +1,29 @@
...
+
+`sync_har.sh chrome 2017-05-15`
+
+This depends on the bucket in Google Storage. If the bucket does not contain a `done` file, the sync is aborted. The file may be set days after the crawl completes.
+
+The script is scheduled to run daily for chrome and android.
+
+### Requests and Pages
+
+The pages and requests tables are generated by running `sync_csv.sh` with the date and an optional `mobile_` prefix:
```
igrigorik: Would it make sense to merge "generating tables" and this section? They're part of the same pipeline, right?

docs/archive.md
rviscomi: It _is_ weird to say that HA archives data on HA. But it's an important piece to distinguish, as the BQ requests/pages tables depend on the CSV data archived here.

docs/archive.md
rviscomi: These are just the data formats. The following sections will go into more detail.

docs/archive.md
rviscomi: I think there should be a distinction between README style docs where we give tips and instructions for how to use HA, and /docs where we describe how the system works. Here, I think it's valuable to be complete and cover the different ways HA persists the data because if not here then where else?

docs/bigquery.md
rviscomi: This is a subsection (###).

docs/archive.md
igrigorik: Right, that's my pedantic point: there is no "SQL" as a data format. MySQL dump is a special flavor. That said, I am nitpicking.

docs/archive.md
igrigorik: Yeah, I guess that's fine. We can hopefully kill off the sql/csv bits very soon (tm).

docs/bigquery.md
igrigorik: Doh, right. Looks good.

docs/archive.md
Themanwithoutaplan: I hope not, I find the CSV dumps very useful.

igrigorik: @Themanwithoutaplan we are moving away from MySQL. Those dumps are a byproduct of the current site, which relies on a custom rollup+processing pipeline to aggregate results... which doesn't lend itself to reprocessing and very hard to update. 

We will continue providing the raw dataset + BigQuery aggregates that should closely match what you see today (and can query via API or export yourself), plus more.. If you haven't already definitely take a look at the BigQuery route.

docs/archive.md
rviscomi: Updated to sqldump

PR #122, opened 2017-07-13
ensure firstHtml is type html

### LINKED ISSUE #120: O(10000) firstHtml requests are certificate checks
```sql
SELECT
  COUNT(0) AS volume,
  resp_content_type
FROM
  [httparchive:runs.2017_06_15_requests]
WHERE
  firstHtml = TRUE
GROUP BY
  resp_content_type
ORDER BY
  volume DESC
```

volume | resp_content_type
-- | --
196243 | text/html; charset=UTF-8
118385 | text/html; charset=utf-8
96333 | text/html
16070 | text/html;charset=UTF-8
**13874** | **application/ocsp-response**
5170 | text/html;charset=utf-8
3823 | text/html; charset=windows-1251
3722 | text/html; charset=ISO-8859-1
2060 | text/html; charset=iso-8859-1
1357 | text/html; charset=gbk
1276 | text/html;charset=ISO-8859-1
1010 | text/html; charset="UTF-8"
640 | text/css
588 | text/html; charset=WINDOWS-1251
585 | text/html; Charset=utf-8
**580** | **application/x-x509-ca-cert**

@pmeenan is this WAI?

rviscomi:
Fixes #120

### DIFF
--- a/bulktest/batch_lib.inc
+++ b/bulktest/batch_lib.inc
@@ -661,9 +661,8 @@ function importEntries($entries, $pageid, $statusInfo, &$firstUrl, &$firstHtmlUr
 			$bFirstReq = 1;
 			$firstUrl = $url;
 		}
-		if ( ! $firstHtmlUrl && 200 == $status ) {
+		if ( ! $firstHtmlUrl && 200 == $status && $type == 'html' ) {
 			// This is the first URL found associated with the page that's HTML.
-			// We could check that mimeType like "%html%", but some HTML responses have the wrong content-type.
 			$bFirstHtml = 1;
 			$firstHtmlUrl = $url;
 		}


bulktest/batch_lib.inc
```diff
@@ -661,9 +661,8 @@ function importEntries($entries, $pageid, $statusInfo, &$firstUrl, &$firstHtmlUr
 			$bFirstReq = 1;
 			$firstUrl = $url;
 		}
-		if ( ! $firstHtmlUrl && 200 == $status ) {
+		if ( ! $firstHtmlUrl && 200 == $status && $type == 'html' ) {
 			// This is the first URL found associated with the page that's HTML.
-			// We could check that mimeType like "%html%", but some HTML responses have the wrong content-type.
```
igrigorik: "but some HTML responses have the wrong content-type"... any chance we can quantify if/how big of a problem this is? Naively.. find pages that have <body> tag but non html content type, or some such?

Also, add a comment to note why we're filtering by HTML?

rviscomi: I found that 96.5% of firstHtml responses already contain `text/html` in their content type header, so this won't make a big difference.

```sql
SELECT
  COUNT(0) AS volume,
  SUM(volume) OVER (ORDER BY volume DESC) AS total,
  RATIO_TO_REPORT(volume) OVER () AS ratio,
  SUM(ratio) OVER (ORDER BY volume DESC) AS total_ratio,
  LOWER(resp_content_type) CONTAINS 'html' AS isHtml
FROM
  [httparchive:runs.2017_06_15_requests]
WHERE
  firstHtml = TRUE 
GROUP BY
  isHtml
ORDER BY
  volume DESC
```


volume | total | ratio | total_ratio | isHtml
-- | -- | -- | -- | --
455013 | 455013 | 0.9649673829 | 0.9649673829 | TRUE
16519 | 471532 | 0.03503261709 | 1 | FALSE

It turns out that counting responses containing `<html...` tag and joining that with responses with/without `text/html` content type is tremendously resource intensive.

```sql
SELECT
  COUNT(DISTINCT(html.url)),
  types.isHtml AS isHtml
FROM (
  SELECT
    url
  FROM
    [httparchive:scratchspace.html_urls]) AS html
JOIN (
  SELECT
    url,
    html AS isHtml
  FROM
    [httparchive:scratchspace.html_types] ) AS types
ON
  types.url = html.url
GROUP BY
  isHtml
```

volume | isHtml
-- | --
436722 | FALSE
1168140 | TRUE

FWIW that query took 20 minutes to run and I had to use a couple of scratch tables.

html_urls:
```sql
SELECT
  url
FROM
  [httparchive:har.2017_06_15_chrome_requests_bodies]
WHERE
  body CONTAINS '<html'
```

html_types:
```sql
SELECT
  url,
  LOWER(resp_content_type) CONTAINS 'html' AS html
FROM
  [httparchive:runs.2017_06_15_requests]
```

 So the results are saying that there were 1.6M distinct requests whose responses included something resembling the `<html>` tag. Of those, 73% also had a Content-Type header containing `html`.

Putting the two together, the remaining 27% are either false positives (not actually HTML) or unlikely to be the canonical HTML document for the page. For example, the most common HTML resource is a Facebook widget, followed by other social/ad widgets. By definition, these can't be `firstHtml` 96.5% of the time.


> Also, add a comment to note why we're filtering by HTML?

A comment seemed necessary when firstHtml wasn't explicitly selecting the first HTML resource, but now that it is, I feel like the code is self explanatory.

rviscomi:
Note to self that this needs to be deployed live.

PR #131, opened 2017-08-31
custom metric: anchor text

### LINKED ISSUE #130: Custom metric: Anchor text
To facilitate research into what kinds of anchor text is popular on the web, add a new custom metric that extracts the `innerText` from each anchor and aggregates them into an array.

If we're able to get this submitted quickly, it can be included in the crawl starting tomorrow.

Related https://github.com/GoogleChrome/lighthouse/issues/3179

rviscomi:
fixes #130

### DIFF
--- a/custom_metrics/anchor_text.js
+++ b/custom_metrics/anchor_text.js
@@ -1,0 +1,1 @@
+return JSON.stringify(Array.from(document.querySelectorAll('a')).map(a => a.innerText.trim()));


rviscomi:
Something to consider: pages may have a ton of anchors and/or very long anchor text, so this could inflate the HAR size by a lot and affect our ability to import them into BQ. We may want to truncate this array in the custom metric.

PR #134, opened 2017-09-12
Remove anchor text custom metric

rviscomi:
Was only included in one crawl (9/1).


rviscomi:
Can merge any time, but should only pull down to the HA server after the 9/1 crawl is complete (or at least done testing) and before the 9/15 crawl begins.

PR #137, opened 2017-10-30
Fixed the request cleanup script

pmeenan:
Fixed the table name for the mobile crawls and eliminated the row counting pass when actually deleting (cuts time in half)


rviscomi:
:+1: thanks Pat!

PR #144, opened 2018-01-23
Removed archive.org donation email

pmeenan:
Use the normal discussion contact path for donation inquiries. The email was getting nothing but spam.


rviscomi:
Should we retain a more private way for potential sponsors to reach out to us, like a donations@ alias that we maintain?

igrigorik:
+1 to donations@ alias.

pmeenan:
I'm a bit split.  Any alias is going to be 99%+ spam and if they post to the forums and want to be more anonymous presumably they'd just post that they were interested in donating and we could exchange private info there.  I assume most companies really doing donations are going to either be solicited or know who to ping directly anyway.

PR #154, opened 2018-08-16
Extend list of mimetypes and extensions for videos

Themanwithoutaplan:
Not sure how to test this. Looks like my editor has stripped trailing whitespace as well.

### DIFF
--- a/bulktest/batch_lib.inc
+++ b/bulktest/batch_lib.inc
@@ -965,7 +965,7 @@ function prettyType($mimeType, $ext) {
 		return "xml";
 	}
     //Video extensions mp4, webm, ts, m4v, m4s, m4v, mov, ogv
-	else if ( false !== strpos($mimeType, "flash") || (true === strpos($mimeType, "webm") || (true === strpos($mimeType, "mp4") || (true === strpos($mimeType, "flv"))
+	else if ( false !== strpos($mimeType, "flash") || (false !== strpos($mimeType, "webm") || (false !== strpos($mimeType, "mp4") || (false !== strpos($mimeType, "flv"))
          "mp4" === $ext || "webm" === $ext || "ts" == $ext || "m4v" === $ext || "m4s" === $ext || "mov" === $ext || "ogv" === $ext ||
          "swf" === $ext || "f4v" === $ext || "flv" === $ext ) {
 		return "video";
--- a/bulktest/batch_lib.inc
+++ b/bulktest/batch_lib.inc
@@ -965,9 +965,9 @@ function prettyType($mimeType, $ext) {
 		return "xml";
 	}
     //Video extensions mp4, webm, ts, m4v, m4s, m4v, mov, ogv
-	else if ( false !== strpos($mimeType, "flash") || (false !== strpos($mimeType, "webm") || (false !== strpos($mimeType, "mp4") || (false !== strpos($mimeType, "flv"))
-         "mp4" === $ext || "webm" === $ext || "ts" == $ext || "m4v" === $ext || "m4s" === $ext || "mov" === $ext || "ogv" === $ext ||
-         "swf" === $ext || "f4v" === $ext || "flv" === $ext ) {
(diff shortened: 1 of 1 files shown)


bulktest/batch_lib.inc
```diff
@@ -964,7 +964,10 @@ function prettyType($mimeType, $ext) {
 	else if ( "xml" === $ext ) {
 		return "xml";
 	}
-	else if ( false !== strpos($mimeType, "flash") || "mp4" === $ext || "swf" === $ext || "f4v" === $ext || "flv" === $ext ) {
+    //Video extensions mp4, webm, ts, m4v, m4s, m4v, mov, ogv
+	else if ( false !== strpos($mimeType, "flash") || (true === strpos($mimeType, "webm") || (true === strpos($mimeType, "mp4") || (true === strpos($mimeType, "flv"))
```
rviscomi: Careful, [`strpos`](http://php.net/manual/en/function.strpos.php) never returns `true`. Let's stick with the syntax used for flash:

```php
false !== strpos($mimeType, "webm")
```

bulktest/batch_lib.inc
```diff
@@ -964,7 +964,10 @@ function prettyType($mimeType, $ext) {
...
+         "mp4" === $ext || "webm" === $ext || "ts" == $ext || "m4v" === $ext || "m4s" === $ext || "mov" === $ext || "ogv" === $ext ||
```
rviscomi: Never seen `ts` before. Turns out it's [used by DVDs](https://fileinfo.com/extension/ts). Is this really something we need to check? Possible collisions with TypeScript files?

bulktest/batch_lib.inc
Themanwithoutaplan: OK. Such a pity we don't have this code in Python…

bulktest/batch_lib.inc
Themanwithoutaplan: ts is the third most common extension according to the current data. It's used in DVB streams and in VOD platforms.

From the query I ran:

Extention | MIMEType | Count 
-----------------------------
mp4 | video/mp4 | 142048
  | video/webm | 11836  
ts | video/MP2T | 11826

Themanwithoutaplan:
Will these changes also be picked in the newer version of the site? We probably should also add video by type to the summary.

bulktest/batch_lib.inc
```diff
@@ -964,7 +964,10 @@ function prettyType($mimeType, $ext) {
 	else if ( "xml" === $ext ) {
 		return "xml";
 	}
-	else if ( false !== strpos($mimeType, "flash") || "mp4" === $ext || "swf" === $ext || "f4v" === $ext || "flv" === $ext ) {
+    //Video extensions mp4, webm, ts, m4v, m4s, m4v, mov, ogv
+	else if ( false !== strpos($mimeType, "flash") || (false !== strpos($mimeType, "webm") || (false !== strpos($mimeType, "mp4") || (false !== strpos($mimeType, "flv"))
```
rviscomi: I think we can remove the paren group after the flash condition. Also, at the end of this lie we need `||` to join with the next line (fixes a syntax error).

rviscomi:
> Will these changes also be picked in the newer version of the site?

Yes, assuming we `git pull` on the crawl server after this change lands. That will ensure the CSVs have the updated video data, which the new site queries in the `summary_pages` tables.

Themanwithoutaplan:
Interestingly I can't find any pages with AV1 examples which should now work in Cr Canary or FF Nightly with the flags set. There is supposed to be a video for Chrome but Jai Krishnan who trailed this at NAB 2018 helpfully forgot to make sure the URL was posted. But currently Cr and FF will happily accept Matroska (WebM or MKV) containers.

Themanwithoutaplan:
I can't see the results of the changes in any of the reports.

rviscomi:
You should see the change in the September 1 crawl.

PR #160, opened 2019-06-24
Almanac custom metrics

ymschaap:
This adds my 'Query Written' items for the Almanac.

cc @rviscomi

### DIFF
--- a/custom_metrics/almanac.js
+++ b/custom_metrics/almanac.js
@@ -19,25 +19,210 @@ return {
   '01.13': document.querySelector('script[nomodule]') ? 1 : 0,
   'link-nodes': (() => {
     // Returns a JSON array of link nodes and their key/value attributes.
-    // Used by 01.14, 01.15, 01.16...
+    // Used by 01.14, 01.15, 01.16, 10.6
     var nodes = document.querySelectorAll('head link');
     var linkNodes = [];
 
     if (nodes) {
-        for (var i = 0, len = nodes.length; i < len; i++) {
-            var node = nodes[i];
-            var attributes = Object.values(node.attributes);
-            var el = {};
+      for (var i = 0, len = nodes.length; i < len; i++) {
+        var node = nodes[i];
+        var attributes = Object.values(node.attributes);
+        var el = {};
 
-            for (var n = 0, len2 = attributes.length; n < len2; n++) {
-                var attribute = attributes[n];
-                el[attribute.name.toLowerCase()] = attribute.value;
-            }
-
-            linkNodes.push(el);
+        for (var n = 0, len2 = attributes.length; n < len2; n++) {
+          var attribute = attributes[n];
+          el[attribute.name.toLowerCase()] = attribute.value;
         }
+
+        linkNodes.push(el);
+      }
     }
-    
+
     return JSON.stringify(linkNodes);
+  })(),
+  'meta-nodes': (() => {
+    // Returns a JSON array of meta nodes and their key/value attributes.
+    // Used by 10.6, 10.7 (potential: 09.29, 12.5, 04.5)
+    var nodes = document.querySelectorAll('head meta');
+    var metaNodes = [];
+
+    if (nodes) {
+      for (var i = 0, len = nodes.length; i < len; i++) {
+        var node = nodes[i];
+        var attributes = Object.values(node.attributes);
+        var el = {};
+
+        for (var n = 0, len2 = attributes.length; n < len2; n++) {
+          var attribute = attributes[n];
+          el[attribute.name.toLowerCase()] = attribute.value;
+        }
+
+        metaNodes.push(el);
+      }
+    }
+
+    return JSON.stringify(metaNodes);
+  })(),
+  // Extract schema.org elements and finds all @context and @type usage
+  '10.5': (() => {
+    var nodes = document.querySelectorAll('[itemtype], script[type=\'application/ld+json\']');
+    var link = document.createElement('a');
+    var schemaElements = {};
+
+    function nestedLookup(items) {
+      var keys = Object.keys(items);
+      for (var i = 0, len = keys.length; i < len; i++) {
+        var item = items[keys[i]];
+        // if array or object, dive into it
+        if (item instanceof Object || item instanceof Array) {
+          nestedLookup(item);
+        }
+      }
+      if (items['@type']) {
+        if (items['@context']) {
+          link.href = items['@context'] + '/' + items['@type'];
+          schemaElements[link.hostname + link.pathname] = true;
+        } else {
+          schemaElements[items['@type']] = true;
+        }
+      }
+    }
+
+    if (nodes) {
+      for (var i = 0, len = nodes.length; i < len; i++) {
+        var node = nodes[i];
+        var item = node.getAttribute('itemtype');
+
+        if (item) {
+          // microdata
+          link.href = node.getAttribute('itemtype');
+          schemaElements[link.hostname + link.pathname] = true;
+        } else if ((node.tagName = 'SCRIPT')) {
+          // json+ld
+          var content = JSON.parse(node.textContent);
+          var contentLoop = [];
+          if (content) {
+            // nested lookup
+            nestedLookup(content);
+          } else {
+            // flag failed json parse?
+          }
+        }
+      }
+    }
+    return JSON.stringify(Object.keys(schemaElements));
+  })(),
+  // Looks at links and identifies internal, external or hashed
+  'seo-anchor-elements': (() => {
+    // metric 10.10, 10.11
+    var nodes = document.getElementsByTagName('a');
+    var link = document.createElement('a');
+
+    var internal = 0; // metric 10.11
+    var external = 0; // metric 10.11
+    var hash = 0;
+    var navigateHash = 0; // metric 10.11
+    var issue = 0;
+
+    if (nodes) {
+      for (var i = 0, len = nodes.length; i < len; i++) {
+        var node = nodes[i];
+
+        // our local parser trick
+        if (node.href) {
+          link.href = node.href;
+
+          if (document.location.hostname === link.hostname) {
+            internal++;
+            if (document.location.pathname === link.pathname && link.hash.length > 0) {
+              // check if hash matches an element in the DOM (scroll to)
+              var element = document.querySelector(link.hash);
+              if (element) {
+                hash++;
+              } else {
+                navigateHash++;
+              }
+            }
+          } else if (document.location.hostname !== link.hostname) {
+            external++;
+          } else {
+            issue++;
+          }
+        }
+      }
+    }
+
+    return JSON.stringify({ internal, external, hash, navigateHash, issue });
+  })(),
+  // Extracts titles used and counts the words, to flag thin content pages
+  'seo-titles': (() => {
+    //metric 10.9
+    var nodes = document.querySelectorAll('h1, h2, h3, h4');
+    var titleWords = -1;
+    var titleElements = -1;
+
+    if (nodes) {
+      titleWords = 0;
+      titleElements = 0;
+
+      // analyse each title node
+      for (var i = 0, len = nodes.length; i < len; i++) {
+        var node = nodes[i];
+        // remove extra whitespace
+        var nodeText = node.textContent.replace(/\s+/g, ' ').replace(/^\s+|\s+$/, '');
+        var nodeWordsCount = nodeText.split(' ').length;
+
+        if (nodeWordsCount > 0) {
+          titleWords += nodeWordsCount;
+          titleElements++;
+        }
+      }
+    }
+    return JSON.stringify({ titleWords, titleElements });
+  })(),
+  // Extracts words on the page to flag thin content pages
+  'seo-words': (() => {
+    //metric 10.9
+    var body = (document.getElementsByTagName('body') || [])[0];
+    var wordsCount = -1;
+    var wordElements = -1;
+    if (body) {
+      wordsCount = 0;
+      wordElements = 0;
+      var n,
+        nodes = [],
+        walk = document.createTreeWalker(
+          body,
+          NodeFilter.SHOW_ALL,
+          {
+            acceptNode: function(node) {
+              if (node.nodeName === 'STYLE' || node.nodeName === 'SCRIPT') {
+                return NodeFilter.FILTER_REJECT;
+              }
+              if (node.nodeType !== 3) {
+                return NodeFilter.FILTER_SKIP;
+              }
+              return NodeFilter.FILTER_ACCEPT;
+            }
+          },
+          false
+        );
+      while ((n = walk.nextNode())) nodes.push(n);
+
+      // analyse each text node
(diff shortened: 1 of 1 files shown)


rviscomi:
Thanks @ymschaap! Great to see so many new metrics being supported.

Could you share the WPT ID of your test so I can verify the JSON results?

ymschaap:
Yes:
WPT ID: 190625_SK_72389e6d83fe6fbf2648977d9062db3d

I must note that returning an object doesn't work for me with the web interface, only if I wrap the results in a JSON.stringify:
```
[almanac]
return JSON.stringify({
  ...
});

https://www.webpagetest.org/custom_metrics.php?test=190625_SK_72389e6d83fe6fbf2648977d9062db3d&run=1&cached=0

In this case this results to double JSON encoding. 

Please check with the WPT team if returning an object as CM works, or if it needs to be a string.


```

rviscomi:
Hmm good catch. Let's always return a stringified object, which should enable each metric to be returned as complex arrays/objects. That also avoids double-JSON encoding. Could you update your metrics to remove the inner stringify?

rviscomi:
I also see a lot of redundant code for some metrics. You should break these out into utility functions.

custom_metrics/almanac.js
```diff
@@ -19,25 +19,210 @@ return {
   '01.13': document.querySelector('script[nomodule]') ? 1 : 0,
   'link-nodes': (() => {
     // Returns a JSON array of link nodes and their key/value attributes.
-    // Used by 01.14, 01.15, 01.16...
+    // Used by 01.14, 01.15, 01.16, 10.6
```
rviscomi: Also 06.46

ymschaap:
* removed double string encoding
* moved duplicate code to function
* added 12.12, 12.14, 06.47 CM

custom_metrics/almanac.js
```diff
@@ -12,32 +12,263 @@
...
+return JSON.stringify({
   // Wether the page contains <script type=module>.
   '01.12': document.querySelector('script[type=module]') ? 1 : 0,
   // Wether the page contains <script nomodule>.
   '01.13': document.querySelector('script[nomodule]') ? 1 : 0,
   'link-nodes': (() => {
     // Returns a JSON array of link nodes and their key/value attributes.
-    // Used by 01.14, 01.15, 01.16...
+    // Used by 01.14, 01.15, 01.16, 10.6,  06.46, 12.18
+    function parseNodes(nodes) {
```
rviscomi: Are you able to pull the function out of the return block so it can be called by multiple metrics? We don't want to define the same function twice because we're still duplicating code.

custom_metrics/almanac.js
```diff
@@ -12,32 +12,263 @@
 // 3. Test your change by following the instructions at https://github.com/HTTPArchive/almanac.httparchive.org/issues/33#issuecomment-502288773.
```
rviscomi: Item 2 above is obsolete now that we're stringifying the whole response. Can you replace that guidance with 2a?

custom_metrics/almanac.js
```diff
@@ -12,32 +12,263 @@
...
+
+        if (nodeWordsCount > 0) {
+          titleWords += nodeWordsCount;
+          titleElements++;
+        }
+      }
+    }
+    return { titleWords, titleElements };
+  })(),
+  // Extracts words on the page to flag thin content pages
```
rviscomi: Is there a significant performance cost to iterating over all nodes in the DOM? For some text-heavy pages, I also wonder if saving everything to an array could eat up significant memory to affect test performance/accuracy.

custom_metrics/almanac.js
ymschaap: I can't see to what code line your remark refers.

rviscomi: I'm unable to comment on a line that you haven't modified. The instruction on the lines above to jsonify objects/arrays is outdated, so it should be removed. If you don't do it in this PR I can submit a follow-up.

custom_metrics/almanac.js
ymschaap: About performance, I'm not sure right person to answer. This is DOM native method.

I've removed the array loop and replaced it with a function call.

custom_metrics/almanac.js
ymschaap: Ok, done.

custom_metrics/almanac.js
```diff
@@ -7,37 +7,223 @@
...
+  })(),
+  // Extract schema.org elements and finds all @context and @type usage
+  '10.5': (() => {
+    function nestedLookup(items) {
+      var keys = Object.keys(items);
+      for (var i = 0, len = keys.length; i < len; i++) {
+        var item = items[keys[i]];
+        // if array or object, dive into it
+        if (item instanceof Object || item instanceof Array) {
+          nestedLookup(item);
```
rviscomi: Hmm it's concerning that this is recursive. We run 5M tests for both desktop and mobile and I worry that O(N^2) algorithms could have an adverse impact on the performance of the crawl. Do you have any anecdotal experience with the size/depth of `items`?

cc @pmeenan for second opinions.

custom_metrics/almanac.js
rviscomi: My concern was mostly about how often the `analyseNode` function would be called per page. Any clue whether it might be called 100 times? 10,000 times? 1,000,000 times?

custom_metrics/almanac.js
```diff
@@ -7,37 +7,223 @@
...
+        var node = nodes[i];
+        var item = node.getAttribute('itemtype');
+
+        if (item) {
+          // microdata
+          link.href = node.getAttribute('itemtype');
+          schemaElements[link.hostname + link.pathname] = true;
+        } else if ((node.tagName = 'SCRIPT')) {
+          // json+ld
+          var content = JSON.parse(node.textContent);
```
rviscomi: We need to ensure that the entire custom metric doesn't fail if the JSON-LD is unable to be parsed. And in general, we should be very careful about building in resiliency because web pages can be... weird.

Can you add a try/catch around this? Any other places that you can think of that could better handle failures?

rviscomi:
Thanks again for working on this. A few more comments. There are quite a few new custom metrics so I'm trying to be diligent.

custom_metrics/almanac.js
ymschaap: analyseNode is only applied to text nodes. I tested on Wikipedia's largest articles:
> https://en.wikipedia.org/wiki/List_of_2018%E2%80%9319_Super_Rugby_transfers = 21,568 times
> https://en.wikipedia.org/wiki/List_of_2018%E2%80%9319_Super_Rugby_transfers = 24770 times

custom_metrics/almanac.js
ymschaap: I understand. I could add a maxDepth. E.g. 5 would be more than enough for parsing schema.org data.

custom_metrics/almanac.js
ymschaap: Good catch! Added it. As well for a querySelector based on _user_ input.

rviscomi:
Is this ready for review?

ymschaap:
Yes. I made the changes.

custom_metrics/almanac.js
```diff
@@ -7,37 +7,233 @@
...
+    if (nodes) {
+      titleWords = 0;
+      titleElements = 0;
 
-            linkNodes.push(el);
+      // analyse each title node
+      for (var i = 0, len = nodes.length; i < len; i++) {
+        var node = nodes[i];
+        // remove extra whitespace
+        var nodeText = node.textContent.replace(/\s+/g, ' ').replace(/^\s+|\s+$/, '');
```
rviscomi: Rather than the second `replace`, use [trim()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/Trim)
```suggestion
        var nodeText = node.textContent.trim().replace(/\s+/g, ' ');
```

custom_metrics/almanac.js
```diff
@@ -7,37 +7,233 @@
...
+        var nodeWordsCount = nodeText.split(' ').length;
+
+        if (nodeWordsCount > 0) {
+          titleWords += nodeWordsCount;
+          titleElements++;
```
rviscomi: If a title has 0 words we won't count it?

custom_metrics/almanac.js
```diff
@@ -7,37 +7,233 @@
...
+      }
+    }
+    return { titleWords, titleElements };
+  })(),
+  // Extracts words on the page to flag thin content pages
+  'seo-words': (() => {
+    //metric 10.9
+    function analyseTextNode(node) {
+      // remove extra whitespace
+      var nodeText = node.textContent.replace(/\s+/g, ' ').replace(/^\s+|\s+$/, '');
```
rviscomi:
```suggestion
      var nodeText = node.textContent.trim().replace(/\s+/g, ' ');
```

custom_metrics/almanac.js
```diff
@@ -7,37 +7,233 @@
...
+      var nodeWordsCount = nodeText.split(' ').length;
+
+      if (nodeWordsCount > 3) {
+        // update counts
+        wordsCount += nodeWordsCount;
+        wordElements++;
+      }
+    }
+
+    var body = document.querySelector('body');
```
rviscomi:
```suggestion
    var body = document.body;
```

custom_metrics/almanac.js
```diff
@@ -7,37 +7,233 @@
...
+        nodes = [],
+        walk = document.createTreeWalker(
+          body,
+          NodeFilter.SHOW_ALL,
+          {
+            acceptNode: function(node) {
+              if (node.nodeName === 'STYLE' || node.nodeName === 'SCRIPT') {
+                return NodeFilter.FILTER_REJECT;
+              }
+              if (node.nodeType !== 3) {
```
rviscomi: What's nodeType 3? Maybe leave a comment explaining what it is.

custom_metrics/almanac.js
```diff
@@ -7,37 +7,233 @@
...
+                return NodeFilter.FILTER_SKIP;
+              }
+              return NodeFilter.FILTER_ACCEPT;
+            }
+          },
+          false
+        );
+      while ((n = walk.nextNode())) analyseTextNode(n);
```
rviscomi: How does this handle nested elements? For example:

```html
<div>
   Foo <span>Bar</span>
</div>
```

Does the div have 2 words and the span has 1? 1 and 1? 2 and 0?

Trying to understand if some text might be double-counted.

custom_metrics/almanac.js
```diff
@@ -7,37 +7,233 @@
...
+    }
+    return { wordsCount, wordElements };
+  })(),
+  // Parse <input> elements
+  'input-elements': (() => {
+    // Used by  12.12, 12.14
+    var nodes = document.querySelectorAll('body input, body select');
```
rviscomi: I don't think the body part of the selector is necessary. input/select can't be found anywhere else (valid, at least).
```suggestion
    var nodes = document.querySelectorAll('input, select');
```

custom_metrics/almanac.js
```diff
@@ -7,37 +7,233 @@
...
+    var inputNodes = parseNodes(nodes);
+
+    return inputNodes;
+  })(),
+  // Find first child of <head>
+  '06.47': (() => {
+    var head = document.querySelector('head');
```
rviscomi:
```suggestion
    var head = document.head;
```

custom_metrics/almanac.js
```diff
@@ -7,37 +7,233 @@
...
+    if (head) {
+      var headChild = head.firstElementChild;
+      if (headChild && headChild.tagName == 'LINK' && headChild && /fonts.googleapis.com/i.test(headChild.getAttribute('href'))) {
```
rviscomi: Redundant condition

```suggestion
      if (headChild && headChild.tagName == 'LINK' && /fonts.googleapis.com/i.test(headChild.getAttribute('href'))) {
```

custom_metrics/almanac.js
```diff
@@ -7,37 +7,233 @@
...
+  })(),
+  // Parse <input> elements
+  'input-elements': (() => {
+    // Used by  12.12, 12.14
+    var nodes = document.querySelectorAll('body input, body select');
+    var inputNodes = parseNodes(nodes);
+
+    return inputNodes;
+  })(),
+  // Find first child of <head>
```
rviscomi:
```suggestion
  // Whether the first child of <head> is a Google Fonts <link>
```

custom_metrics/almanac.js
```diff
@@ -7,37 +7,233 @@
...
+    if (nodes) {
+      for (var i = 0, len = nodes.length; i < len; i++) {
+        var node = nodes[i];
+        var item = node.getAttribute('itemtype');
+
+        if (item) {
+          // microdata
+          link.href = node.getAttribute('itemtype');
+          schemaElements[link.hostname + link.pathname] = true;
+        } else if ((node.tagName = 'SCRIPT')) {
```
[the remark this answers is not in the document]
ymschaap: 🤭

custom_metrics/almanac.js
```diff
@@ -7,37 +7,233 @@
...
+        var item = items[keys[i]];
+        // if array or object, dive into it
+        if (item instanceof Object || item instanceof Array) {
+          nestedLookup(item, depth++);
+        }
+      }
+      if (items['@type']) {
+        if (items['@context']) {
+          link.href = items['@context'] + '/' + items['@type'];
+          schemaElements[link.hostname + link.pathname] = true;
```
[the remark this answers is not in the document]
ymschaap: Assigning user input to the `a` anchor href, cleans up very well from my experience. I tested, and if any of the items are (unexpected) function, array, object, non-western characters it won't fail (and would encode it to a string, and set that as pathname).

If @context is not a valid hostname e.g. "/schema.org" or "schema", it will also not fail (but have link.hostname return as the document.location.hostname)

We could check if @context starts with 'http(s)://' and/or  `typeof items[@type'] === "string"` to be on the save side?

I'm open to any alternatives, given I have no experience with parsing millions of documents.

custom_metrics/almanac.js
ymschaap: Please check your console.log for this fiddle:
> https://jsfiddle.net/3c9z046p/

1. yes, foo and bar are two nodes in this case, and would have each one word.
2. the alternative is not more reliable imho (e.g. by looking only into the text of block nodes (p, div, li).
3. it won't be double counted.

If there is a better way, let me know.

custom_metrics/almanac.js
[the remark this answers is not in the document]
ymschaap: This is a test how a.href handles user input.
> https://jsfiddle.net/3sd7m6ew/

Check your console.log.

ymschaap:
Latest changes ready.

rviscomi:
Huge thanks for slogging through my feedback @ymschaap! Really looking forward to getting these new metrics in the crawl.

PR #162, opened 2019-06-28
feat: add link-preconnect

tjmonsi:
This adds a list of URLs that was fetched using `link rel="preconnect"` for aggregation.

@rviscomi please check

### DIFF
--- a/custom_metrics/almanac.js
+++ b/custom_metrics/almanac.js
@@ -239,5 +239,21 @@ return JSON.stringify({
       }
     }
     return 0;
+  })(),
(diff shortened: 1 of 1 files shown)


rviscomi:
Sorry for the false alarm, but this metric isn't actually necessary because https://github.com/HTTPArchive/legacy.httparchive.org/blob/master/custom_metrics/almanac.js#L39 should do exactly what you're looking for. Tried editing my [comment](https://github.com/HTTPArchive/almanac.httparchive.org/issues/33#issuecomment-506753094) in time.

PR #167, opened 2019-09-15
Properly detect SVG clickables with no text

obto:
This is the PR we discussed last week.

Fixes the issue where SVG clickables of any kind were being detected

### DIFF
--- a/custom_metrics/almanac.js
+++ b/custom_metrics/almanac.js
@@ -278,19 +278,42 @@ return JSON.stringify({
     return parsedNodes;
   })(),
   '12.11': (() => {
+    function containsAnSvg(element) {
+      var children = Array.from(element.childNodes);
+      return !!children.find((child) => {
+        if (child.tagName && child.tagName.toLowerCase() === 'svg') {
+          return true;
+        }
+
+        if (child.childNodes.length) {
+          return containsAnSvg(child);
+        }
+
+        return false;
+      });
+    }
+
     // Counts the links or buttons only containing an icon.
     var clickables = document.querySelectorAll('a, button');
     return Array.from(clickables).reduce((n, clickable) => {
-      // Clickables containing SVG are assumed to be icons.
-      if (clickable.firstElementChild && clickable.firstElementChild.tagName == 'SVG') {
-        return n + 1;
-      }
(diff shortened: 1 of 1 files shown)


obto:
@rviscomi Is it possible to merge this before the next run?

rviscomi:
Sorry this slipped past me. The October crawl started a couple days ago, but it wouldn't be completed until the end of the month anyway. That wouldn't leave us much time to get the results analyzed and updated in the chapter before launch. :/

obto:
Hmm since the detection of links using font icons was accurate I think I'll do the following to get a rough estimate for the Almanac then:

1. Run an analysis to find out what % of pages (which have buttons and links) have any svgs on them at all
2. Remove that % from the metric result we already have... which should effectively give me the % of pages using clickables with only a font icon

This should give me a pretty good lower bound, and I can go back and edit the Almanac at a later point when we re-run it in the future.

rviscomi:
SGTM 👍

obto:
Great :+1: 

Is this PR alright to merge then?

rviscomi:
It is, but there's another step which is for me to sync it with the HTTP Archive test server. I'll hold off on merging until I have time to sync before the next crawl as a reminder.

rviscomi:
Wow I had forgotten this change existed! I'll merge now to get it in before the August crawl. Sorry @obto! 😅

custom_metrics/almanac.js
```diff
@@ -278,19 +278,42 @@ return JSON.stringify({
     return parsedNodes;
   })(),
   '12.11': (() => {
+    function containsAnSvg(element) {
+      var children = Array.from(element.childNodes);
+      return !!children.find((child) => {
```
rviscomi, line 283: Actually could this be simplified to

```js
function containsAnSvg(element) {
  return !!element.querySelector('svg');
}
```

obto:
@rviscomi No worries. I've included this change in the branch I'm working on right now for migrating as many responsebody queries into custom metrics. So we can close this PR request.

custom_metrics/almanac.js
obto: Good call. I'll update it in my other branch

PR #171, opened 2020-07-21
Custom metrics for SEO and WPT Bodies

Tiggerito:
### Testing

### Code Architecture
- [x] **try catch logging system**
I've implemented some helper functions to enable logging errors to a log property of the almanac. Using this and try catch blocks around property code will protect us from a one error all lost situation. It will also enable us to query the errors that happen so we can fix them for the next run.

- [x] **$WPT_BODIES**
This is present in the wpt_metrics.js file

The SEO chapter wanted to see both raw and rendered versions of a page. This can be accessed via the special $WPT_BODIES string. I've implemented some helper functions to make it easier to use:

getRawHtml() - string of the raw html
getRawHtmlDocument() - an html document of the raw html
getRawHtmlDiv() - the raw html placed in a div. This was needed for a special case where I need to test for content visibility. I temporarily add the div to the rendered page so that it can calculate things. Note that it seems the head is removed from this version, so it does not work for gathering meta data etc.
getResponseHeaders(name) - returns an array of matching response headers.

### [SEO](https://github.com/HTTPArchive/almanac.httparchive.org/issues/908)
- [x] **anchors**
analysis of link types
rel attributes and the detection of img only links
target _blank checks in relation to noopener and noreferrer
target values used
- [x] **title**
  data about title tags
- [x] **meta_description**
  data about meta description tags
- [x] **hreflang**
get hreflang from link tags and http headers
- [x] **headings**
  data on hX tags
- [x] **structured_data**
  json-ld count and errors
  types used, vocabularies used, formats used
id usage
- [x] **raw_html**
check for missing html, body and head tags
html comments and the use of 'if'
head length
- [x] **canonicals**
canonical data from link tags and header
- [x] **robots**
robots settings from meta tags and http headers. 
- [x] **visible_words**
  count of words visible in the page

Progress on https://github.com/HTTPArchive/almanac.httparchive.org/issues/908


rviscomi:
@Tiggerito I've edited your PR comment to help clarify what's implemented and ready for review. Can you add the rest of the metrics to the checklist that need to be implemented in this PR and check off any that are ready for review? Set the entire PR to "ready for review" when you'd like me to look.

Tiggerito:
I feel like I've already created a beast, and there is more. I'm concerned about getting this completed and reasonably tested in time.

I've actually been trying to reduce the data output today, now I know more on the requirements. e.g. only numbers where possible.

The largest part of the output is the link tag array, which is used for quite a few things. It was there last year which makes me think we can cope with that. Especially when this data is a tiny bit of the total payload.

Something I added today was the use of WebPageSpeeds $WPT_REQUESTS that gives me access to the pages headers. That makes quite a few of my tasks simpler. Is that OK?

And I added a try catch around it all to capture any exceptions. I may now have a look at the errors I saw from last year.

I presume the main concern is the size of the data output, and if the script has a high cpu/time use. The size of the actual script I guess is not that important?

Should I say its ready for review now, or wait till I've tested more and implemented more. Some I'm reasonably confident are stable now, I've checked them off.

Tiggerito:
Current plan is for me to rebase when obto has merged.

Tiggerito:
My stuff is ready. I'm trying to get the team to give me more test cases.

rviscomi:
> Current plan is for me to rebase when obto has merged.

Not ideal but you could also branch from `obto:new_metrics_and_bug_fixes` to get a head start on the merge. I like the idea you mentioned in Slack to save the new functions to seo/markup-specific files. This would help mitigate any risk of functions breaking each other or timing out the custom metric.

> Something I added today was the use of WebPageSpeeds $WPT_REQUESTS that gives me access to the pages headers. That makes quite a few of my tasks simpler. Is that OK?

Could you check that any metrics that makes use of $WPT variables are in their own js file?

> My stuff is ready. I'm trying to get the team to give me more test cases.

Meanwhile could you test the same [URLs](https://docs.google.com/spreadsheets/d/1FliIat7uXauPVCpWcl1GU-InAOWXW7e7HHJuv4c7URQ/edit?usp=sharing) as @obto and add the links to the WPT results in the top PR comment?

Tiggerito:
> > Current plan is for me to rebase when obto has merged.
> 
> Not ideal but you could also branch from `obto:new_metrics_and_bug_fixes` to get a head start on the merge. I like the idea you mentioned in Slack to save the new functions to seo/markup-specific files. This would help mitigate any risk of functions breaking each other or timing out the custom metric.
> 
> > Something I added today was the use of WebPageSpeeds $WPT_REQUESTS that gives me access to the pages headers. That makes quite a few of my tasks simpler. Is that OK?
> 
> Could you check that any metrics that makes use of $WPT variables are in their own js file?
> 
> > My stuff is ready. I'm trying to get the team to give me more test cases.
> 
> Meanwhile could you test the same [URLs](https://docs.google.com/spreadsheets/d/1FliIat7uXauPVCpWcl1GU-InAOWXW7e7HHJuv4c7URQ/edit?usp=sharing) as @obto and add the links to the WPT results in the top PR comment?

I've moved all my code into separate js files, so we should no longer have a merge conflict. Will test then commit the latest version.

Tiggerito:
> > > Current plan is for me to rebase when obto has merged.
> > 
> > 
> > Not ideal but you could also branch from `obto:new_metrics_and_bug_fixes` to get a head start on the merge. I like the idea you mentioned in Slack to save the new functions to seo/markup-specific files. This would help mitigate any risk of functions breaking each other or timing out the custom metric.
> > > Something I added today was the use of WebPageSpeeds $WPT_REQUESTS that gives me access to the pages headers. That makes quite a few of my tasks simpler. Is that OK?
> > 
> > 
> > Could you check that any metrics that makes use of $WPT variables are in their own js file?
> > > My stuff is ready. I'm trying to get the team to give me more test cases.
> > 
> > 
> > Meanwhile could you test the same [URLs](https://docs.google.com/spreadsheets/d/1FliIat7uXauPVCpWcl1GU-InAOWXW7e7HHJuv4c7URQ/edit?usp=sharing) as @obto and add the links to the WPT results in the top PR comment?
> 
> I've moved all my code into separate js files, so we should no longer have a merge conflict. Will test then commit the latest version.

@rviscomi 

It's now based on three custom metrics files:

wpt_metrics.js
markup.js
robots_txt.js

Not sure how to back out all my edits to the almanac.js file? No changes needed there anymore.

Test results are here:

https://docs.google.com/spreadsheets/d/15JiUZ23IGi1covK_7dAEd2bFvAGGirEPxVf70vlgf2k/edit#gid=0

My stuff typically returns 10k if data.

rviscomi:
> Not sure how to back out all my edits to the almanac.js file? No changes needed there anymore.

Try this https://ohshitgit.com/#undo-a-file

> It's now based on three custom metrics files:
>
>wpt_metrics.js
>markup.js
>robots_txt.js

Regarding `wpt_metrics.js`, is there a more specific name we could use like `seo.js`? Or are all of these metrics dependent on `$WPT_BODIES`? If not, can you split out only the bodies-dependent metrics into their own compartmentalized file named `wpt_bodies.js` or similar? My hope is that any failures due to gigantic pages/responses would only affect metrics that depend directly on it.

~2k lines of code is astonishing but at the same time I'm wary about the amount of duplication between this and #172. To what extent do your new metrics cover the same ground?

It's going to take a while to review all of this, unfortunately, but I'll try.

Tiggerito:
> > Not sure how to back out all my edits to the almanac.js file? No changes needed there anymore.
> 
> Try this https://ohshitgit.com/#undo-a-file
> 
> > It's now based on three custom metrics files:
> > wpt_metrics.js
> > markup.js
> > robots_txt.js
> 
> Regarding `wpt_metrics.js`, is there a more specific name we could use like `seo.js`? Or are all of these metrics dependent on `$WPT_BODIES`? If not, can you split out only the bodies-dependent metrics into their own compartmentalized file named `wpt_bodies.js` or similar? My hope is that any failures due to gigantic pages/responses would only affect metrics that depend directly on it.

All in there are dependent on $WPT_BODIES. Not sure why I called it wpt_metrics. Will change. The plan was to isolate the use of $WPT_BODIES as well as let others take advantage of it. 

It just so happens that most of the SEO specific stuff needs it. Some crossover stuff is fine in the markup.js.

If it fails the SEO chapter are mostly back to last years metrics. So I don't want those properties removed from almanac.js just yet. 

> 
> ~2k lines of code is astonishing but at the same time I'm wary about the amount of duplication between this and #172. To what extent do your new metrics cover the same ground?

Take into account I've added a lot of exception handling. And 2k lines is minuscule to what $WPT_BODIES adds.

The exception handling and the fact it is in my own script means any bugs will only affect one property. Unless $WPT_BODIES itself causes a failure. 

There are a couple of large properties. e.g. the Structured Data parser. Again, if that fails, it's only our loss. That would be something not possible to process outside if these scripts. The old almanac.js version is not very good.

Some stuff could be worked out from the meta tags and links dump. That's my backup. The js in BigQuery is not that good and I suspect I will not be able to do as detailed an analysis within the query. I'm trying to work things out at the source.

SEO also wants raw/rendered comparisons. I've structured the code to do a lot of reuse. 

> 
> It's going to take a while to review all of this, unfortunately, but I'll try.

It will, that's why I was getting worried a while back.

obto:
@Tiggerito I know you've probably said elsewhere, but can you refresh my memory on why you need to run the metrics on the html before javascript has modified it?

Google's crawler has been fully updated to use the latest version of chrome and render javascript just like clients. Is this specifically for other search engines which may not be?

Tiggerito:
> @Tiggerito I know you've probably said elsewhere, but can you refresh my memory on why you need to run the metrics on the html before javascript has modified it?
> 
> Google's crawler has been fully updated to use the latest version of chrome and render javascript just like clients. Is this specifically for other search engines which may not be?

SEO is very interested in what is changed from raw html to rendered. 

There's a few key cases:

Many crawlers do not render, SEO wants to know what they see.

Some raw values like the canonical link and robots meta tag can cause a page to not get indexed, even if the rendered one has valid values. e.g. the raw html says noindex, mans Google bot will not bother rendering. An invalid canonical is like a redirect, so the page does not get crawled.

There's a lot of interest in how people are adding structured data via JavaScript.

Fancy JavaScript frameworks often make mistakes with using links that are not crawlable. A raw/rendered comparison may give us a clue to those sorts of errors.

Markup want's to see markup styles and errors. That I put into the too hard bucket.

rviscomi:
One more request to help streamline this PR. Could you break it up into separate PRs for each file? I expect the 1300 line file to take a few iterations but the others should be quick reviews that we can merge sooner. 🙏

Tiggerito:
> One more request to help streamline this PR. Could you break it up into separate PRs for each file? I expect the 1300 line file to take a few iterations but the others should be quick reviews that we can merge sooner. 🙏

I'll have a go tomorrow. Not sure how to fork the same thing multiple times. I'll keep this one as the wpt_pages one.

rviscomi:
https://html.spec.whatwg.org/ is a good test case to include. It contains 200,000+ DOM elements. In [this test](https://www.webpagetest.org/custom_metrics.php?test=200728_66_1981765ac0e889540f429193f04ce623&run=2&cached=0) I'm not seeing the "raw" metrics included anywhere, so this may indicate a bug either in WPT or the custom metric. Maybe the document is too big to handle in the custom metric. cc @pmeenan

Tiggerito:
> > One more request to help streamline this PR. Could you break it up into separate PRs for each file? I expect the 1300 line file to take a few iterations but the others should be quick reviews that we can merge sooner. 🙏
> 
> I'll have a go tomorrow. Not sure how to fork the same thing multiple times. I'll keep this one as the wpt_pages one.

As you saw, I've split them out. Creating branches, deleting files and doing new pulls worked.

pmeenan:
If you use https://www.webpagetest.org/?debug=1 to run the tests it will include a raw debug log from the agent including the devtools commands to run the custom metrics (and any handled exceptions).

Tiggerito:
> https://html.spec.whatwg.org/ is a good test case to include. It contains 200,000+ DOM elements. In [this test](https://www.webpagetest.org/custom_metrics.php?test=200728_66_1981765ac0e889540f429193f04ce623&run=2&cached=0) I'm not seeing the "raw" metrics included anywhere, so this may indicate a bug either in WPT or the custom metric. Maybe the document is too big to handle in the custom metric. cc @pmeenan

Good catch. I'm glad the rest of the script works. It implies I'll still have some data if the WPT_BODIES fails.

My error logging did not report anything, implying WPT_BODIES was valid json but did not have the data I use.

It returned the response and the headers as I still reported on http_status_code.

But it did not contain the response_body, it is undefined.

So I guess a bug or defensive measure for a large response_body.

It makes me more confident that this script will still work if WPT_BODIES plays up.

That page reported a lot of jump to links. 42,675. I had to look, and its a massive index page, with lots of jump to links. Code works 🥳

Tiggerito:
> If you use https://www.webpagetest.org/?debug=1 to run the tests it will include a raw debug log from the agent including the devtools commands to run the custom metrics (and any handled exceptions).

Where is the log output?

pmeenan:
In the main results page to the left of the waterfall for each run will be a link for the "debug log" (next to the timeline and trace links).

Tiggerito:
There are still some of last years queries that have not been converted to using the custom metrics because they required regex on the raw html. I think simple checks for the presence of a pattern. Should I add them?

rviscomi:
Do you have examples?

Tiggerito:
> Do you have examples?

I saw them from the comments for https://github.com/HTTPArchive/legacy.httparchive.org/pull/172

e.g. 01_18.sql

```
    COUNTIF(body LIKE '%Atomics.%') AS atomics,
    COUNTIF(body LIKE '%new Intl.%') AS intl,
    COUNTIF(body LIKE '%new Proxy%') AS proxy,
    COUNTIF(body LIKE '%new SharedArrayBuffer(%') AS sharedarraybuffer,
    COUNTIF(body LIKE '%new WeakMap%') AS weakmap,
    COUNTIF(body LIKE '%new WeakSet%') AS weakset
```

Could become something like:

"atomics": !!getRawHtml()?.match(/Atomics./g),

Needs testing and I would probably place them in the raw_html property where I have a few more like that.

Tiggerito:
> 
> "atomics": !!getRawHtml()?.match(/Atomics./g),
> 
> Needs testing and I would probably place them in the raw_html property where I have a few more like that.

More like in raw_html:

result.atomics = !!rawHtml.match(/Atomics\./g);

Tiggerito:
This code does not error. I presume it would detect what they looked for last year. I'll hold on in adding it to a commit:

 ```
         // 01_18.sql
          result.apis.atomics = !!rawHtml.match(/Atomics\./g);
          result.apis.intl = !!rawHtml.match(/new Intl\./g);
          result.apis.proxy = !!rawHtml.match(/new Proxy/g);
          result.apis.sharedarraybuffer = !!rawHtml.match(/new SharedArrayBuffer\(/g);
          result.apis.weakmap = !!rawHtml.match(/new WeakMap/g);
          result.apis.weakset = !!rawHtml.match(/new WeakSet/g);

          // 01_24.sql
          result.dynamic_imports = !!rawHtml.match(/import\(/g);

          // 01_25.sql
          result.sourcemaps = !!rawHtml.match(/sourceMappingURL/g);
          
          // 02_11.sql
          result.reset_util_in_stylesheet = !!rawHtml.match(/normalize\.css|pure\-css|http:\/\/meyerweb\.com\/eric\/tools\/css\/reset\//g);

          // 02_45.sql - Captures class atrributes. I think captured by the dom is better

          // 11_06.sql
          result.beforeinstallprompt = !!rawHtml.match(/beforeinstallprompt/g);
```

**Comment by Tiggerito** [comment]
> In the main results page to the left of the waterfall for each run will be a link for the "debug log" (next to the timeline and trace links).

Still can't get it. No worries.

**Comment by rviscomi** [comment]
> This code does not error. I presume it would detect what they looked for last year. I'll hold on in adding it to a commit:

Let's leave it out. These script patterns are unlikely to exist in the HTML file itself and we would need to check all JS bodies as well.

**Comment by Tiggerito** [comment]
> > This code does not error. I presume it would detect what they looked for last year. I'll hold on in adding it to a commit:
> 
> Let's leave it out. These script patterns are unlikely to exist in the HTML file itself and we would need to check all JS bodies as well.

Ah good point. I forgot the bodies table include resources. A possible for the future.

File: custom_metrics/wpt_bodies.js [hunk 1]
```diff
@@ -0,0 +1,1366 @@
...
+    'title': (() => {
+      try {    
+        let result = {};
+
+        function getTitles(d) {
+          let target = {};
+          target.total = Array.from(d.querySelectorAll('head title')).map(e => {
+            let text = e.innerText.trim();
+            let characters = text.length;
+            let words = text.match(/\S+/g)?.length;
```
rviscomi, line 519: Curious, how does this behave for non-English languages or non-Latin alphabets?

Similarly for `characters`, could it be double-counting 2-byte characters?

custom_metrics/wpt_bodies.js
```diff
@@ -0,0 +1,1366 @@
...
+        let result = {};
+
+        function getMetaDescriptions(d) {
+          let target = {all: {text: "", words: 0, characters: 0}};       
+          target.total = Array.from(d.querySelectorAll('head meta[name="description"]')).map(e => {
+            let text = e.getAttribute("content") ?? "";
+            let characters = text.length;
+            let words = text.match(/\S+/g)?.length;
+
+            target.all.text = (target.all.text+" "+text).trim();
```
rviscomi, line 578: Nit:
```suggestion
            target.all.text += ` ${text.trim()}`;
```

custom_metrics/wpt_bodies.js
```diff
@@ -0,0 +1,1366 @@
...
+        let rawHtmlDocument = getRawHtmlDocument();
+
+        if (rawHtmlDocument) {
+          result.raw = getHreflangValues(rawHtmlDocument);
+        }
+
+        let linkHeaders = getResponseHeaders("link");
+        if (linkHeaders) {
+          linkHeaders.forEach((h) => {
+            let matches = h.matchAll(/hreflang=['"]?(.*?)['"]/gi);
```
rviscomi, line 650: The opening quote is optional but the closing quote isn't?

custom_metrics/wpt_bodies.js
```diff
@@ -0,0 +1,1366 @@
...
+              // looks like a hidden heading
+              target.first_non_empty_heading_hidden = true;
+            }
+          }
+          return {characters: text.length, words: words ?? 0};
+        }
+        
+        function processHeadings(d) {
+          let target = {first_non_empty_heading_hidden: false};
+          for(let l=1; l < 9; l++) {
```
rviscomi, line 711: H6 is the lowest level per https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Heading_Elements. Why go to 8?

custom_metrics/wpt_bodies.js
```diff
@@ -0,0 +1,1366 @@
...
+            let nodes = Array.from(d.querySelectorAll('h'+l));
+
+            let characters = 0;
+            let words = 0;
+            // if don't have a primary heading yet, search for one.
+            
+            var hs = nodes.map(n => {
+              let h = processHeading(d, target, n, l);
+              characters += h.characters;
```
rviscomi, line 720: How useful is it to return a sum of all headings for a level combined? You could still derive the average length, but not the distribution/median for example.

custom_metrics/wpt_bodies.js
```diff
@@ -0,0 +1,1366 @@
...
+    // Used by: SEO
+    // Backup: old data in almanac.js - not very good
+    'structured_data': (() => {  
+      try { 
+        var link = document.createElement('a');
+
+        function nestedJsonldLookup(target, jsonldIds, items, depth, context) {
+          if (items instanceof Array) {
+            // loop array and process any objects in it 
+            for (var i = 0, len = items.length; i < len; i++) {
```
rviscomi, line 764: `items.forEach`?

custom_metrics/wpt_bodies.js
```diff
@@ -0,0 +1,1366 @@
...
+                    target.same_as_values.push(s);
+                  }
+                })
+              }             
+            }
+            
+            if (items['@id']) {
+              if (typeof items['@id'] === 'string') {
+                
+                link.href = items['@id'];
```
rviscomi, line 796: Have you considered using `new URL()`?

custom_metrics/wpt_bodies.js
```diff
@@ -0,0 +1,1366 @@
...
+          // json-ld
+          let jsonld_scripts = Array.from(d.querySelectorAll('script[type="application/ld+json"]'));
+
+          target.jsonld_scripts = {
+            count: jsonld_scripts.length,
+            errors: jsonld_scripts.filter(e => {
+              try {
+                var cleanText = e.textContent.trim();
+                var cleanText = cleanText.replace(/^\/\*(.*?)\*\//g, ''); // remove * comment from start (could be for CDATA section) does not deal with multi line comments
+                var cleanText = cleanText.replace(/\/\*(.*?)\*\/$/g, ''); // remove * comment from end (could be for CDATA section) does not deal with multi line comments
```
rviscomi, line 951: Not clear if this repetition was intentional. I suspect not since the variable is declared three times as well.

custom_metrics/wpt_bodies.js
```diff
@@ -0,0 +1,1366 @@
...
+              }
+              catch(e) {
+                return true; // its bad
+              }
+            }).length
+          };
+
+          // microdata
+          let microdataNodes = d.querySelectorAll('[itemtype]');
+          for (let i = 0, len = microdataNodes.length; i < len; i++) {
```
rviscomi, line 964: `Array.from(microdataNodes).forEach()`?

custom_metrics/wpt_bodies.js
```diff
@@ -0,0 +1,1366 @@
...
+          if (d.querySelector("[itemtype$='SearchAction']")){
+            target.sitelinks_search_box = true;
+          }
+
+          let microdataIds = {};
+
+          let microdataItemIdNodes = [...d.querySelectorAll('[itemid]')];
+
+          microdataItemIdNodes.forEach((n) => {
+            link.href = n.getAttribute('itemid');
```
rviscomi, line 989: The whitespace here is weird

custom_metrics/wpt_bodies.js
```diff
@@ -0,0 +1,1366 @@
...
+      try {
+        let result = {};
+
+        let rawHtml = getRawHtml();
+
+        if (rawHtml) {
+
+          result.body = !!rawHtml.match(/<body/g);
+          result.html = !!rawHtml.match(/<html/g);
+          result.head = !!rawHtml.match(/<head/g);
```
rviscomi, line 1066: Case insensitive?

custom_metrics/wpt_bodies.js
```diff
@@ -0,0 +1,1366 @@
...
+    })(),
+    // canonicals used in the page and http header
+    // Used by: SEO
+    // Backup: link nodes in almanac.js
+    'canonicals': (() => {
+      try {
+        function processCanonical(c) {
+          if (c === result.url) 
+            result.self_canonical = true;
+        else
```
rviscomi, line 1108: Weird whitespace

custom_metrics/wpt_bodies.js
```diff
@@ -0,0 +1,1366 @@
...
+    'canonicals': (() => {
+      try {
+        function processCanonical(c) {
+          if (c === result.url) 
+            result.self_canonical = true;
+        else 
+            result.other_canonical = true;
+
+          if (result.canonicals[c]) 
+            result.canonicals[c]++;
```
rviscomi, line 1112: Can you double check this?

In [this test](https://www.webpagetest.org/custom_metrics.php?test=200729_1R_825019eac58a7c8f83bcf3fb343402f1&run=1&cached=0) I'm getting `canonicals: {https://almanac.httparchive.org/en/2019/: 2}`. And in the [page source](https://almanac.httparchive.org/en/2019/) we have:

```html
<link rel="canonical" href="https://almanac.httparchive.org/en/2019/" />
<link rel="alternate" href="https://almanac.httparchive.org/en/2019/" hreflang="en" />
<link rel="alternate" href="https://almanac.httparchive.org/es/2019/" hreflang="es" />
<link rel="alternate" href="https://almanac.httparchive.org/fr/2019/" hreflang="fr" />
<link rel="alternate" href="https://almanac.httparchive.org/ja/2019/" hreflang="ja" />
<link rel="alternate" href="https://almanac.httparchive.org/en/2019/" hreflang="x-default" />
```

So shouldn't the result be `canonicals: {https://almanac.httparchive.org/en/2019/: 1}`?

custom_metrics/wpt_bodies.js
```diff
@@ -0,0 +1,1366 @@
...
+        let result = {rendered: {}, raw: {}, self_canonical: false, other_canonical: false, canonicals: {}};
+        result.url = document.location.href.split("#")[0];
+
+        // headers
+        let canonicalLinkHeaders = [];
+          // Link: <https://example.com/page-b>; rel="canonical"
+        let linkHeaders = getResponseHeaders("link");
+        if (linkHeaders) {
+          linkHeaders.forEach((h) => {
+            let matches = h.matchAll(/<([^>]*)> *; *rel=['"]?canonical['"]?/gi);
```
rviscomi, line 1126: Should these space characters (` *`) be explicit `\s*`? And do you want to ensure at least one non-`>` character?
```suggestion
            let matches = h.matchAll(/<([^>]+)>\s*;\s*rel=['"]?canonical['"]?/gi);
```

custom_metrics/wpt_bodies.js
Tiggerito: I copied that trick from last years code. With the URL class I'd need to pass in the base URL and test that it can work with all the weird things people enter. Too risky at this stage.

I'll put a comment in the code.

custom_metrics/wpt_bodies.js
Tiggerito: They are slightly different. The extra vars should not be there. Good spot. Updating.

custom_metrics/wpt_bodies.js
Tiggerito: As before. Refactor later.

custom_metrics/wpt_bodies.js
Tiggerito: Good one.

custom_metrics/wpt_bodies.js
Tiggerito: as in it was not tabbed correctly? Fixed that.

custom_metrics/wpt_bodies.js
Tiggerito: I was thinking about it not being a normal space character. Then got it. The lines after are tabbed in one more level for some reason. Fixed.

custom_metrics/wpt_bodies.js
Tiggerito: It's an anomaly of checking bot raw and rendered pages. It's the same tag, just spotted twice. 

I could not decide how to handle it. Maybe just report on what canonicals are found, and forget trying to count them? I'll do that.

Tiggerito:
@rviscomi I seemed to have reviewed my own code. Anyhow, changes/comments made, will test soon.

Tiggerito:
Testing all came out good. One change with how I dealt with the raw title check. I now check if the raw document is present as it is possible it is present but the title is not.

Tiggerito:
> Epic

I'll go with 1, but it's probably 2 😜

image

custom_metrics/wpt_bodies.js
rviscomi: Ah ok, updated my suggestion. It's not the `text` var that you necessarily want to trim, but the concatenated value.

custom_metrics/wpt_bodies.js
rviscomi: 🤷  ok

custom_metrics/wpt_bodies.js
rviscomi: Yes, there's definitely a lot of "data waste" for metrics that don't show remarkable results. Up to you how to aggregate these.

custom_metrics/wpt_bodies.js
rviscomi: Performance is mostly negligible, I suggested for consistency and simplicity, but up to you to resolve as you see fit.

custom_metrics/wpt_bodies.js
Tiggerito:
```
let test = "previous text";
let text = " new text ";
test += ` ${text}`.trim();
console.log(test); // result = 'previous textnew text'
```

rviscomi:
@Tiggerito is this ready for review? Hoping to get this merged tonight 😬

custom_metrics/wpt_bodies.js
Tiggerito: I bit the bullet and refactored forEach and array related stuff.

custom_metrics/wpt_bodies.js
Tiggerito: If it's ok with you, can we leave the waste in for this year?

Tiggerito:
@rviscomi ready for hopefully final review.

Tiggerito:
Doh, I forgot to commit. Done now. I also added an extra trim() to that string thing to cover all bases. Was considering using an array join, but that's over the top.

custom_metrics/wpt_bodies.js
Tiggerito: It's now:
target.all.text = (target.all.text+" "+text.trim()).trim();

custom_metrics/wpt_bodies.js
rviscomi: SGTM

Tiggerito:
> Thanks for helping to get this in shape!

Thanks for the merge. All done now. I'm taking the afternoon off. 🥳

PR #174, opened 2020-07-27
Add custom metrics for Security chapter

tomvangoethem:
This adds the following metrics:

* `allow` and `sandbox` attribute for `<iframe>` elements
* Elements with an `integrity` attribute

Upon merge, the '08.39' metric in almanac.js can be removed (superseded by the `sri-integrity` metric)

Progress on https://github.com/HTTPArchive/almanac.httparchive.org/issues/906

### DIFF
--- a/custom_metrics/security.js
+++ b/custom_metrics/security.js
@@ -1,21 +1,20 @@
 return JSON.stringify({
-	"iframe-allow": [].map.call(document.querySelectorAll('iframe[allow]'), (x) => {
-		return {
-			"allow": x.getAttribute('allow'),
-			"src": x.getAttribute(src)
-		}
-	}),
-	"iframe-sandbox": [].map.call(document.querySelectorAll('iframe[sandbox]'), (x) => {
-		return {
-			"sandbox": x.getAttribute('sandbox'),
-			"src": x.getAttribute(src)
-		}
-	}),
-	"sri-integrity": [].map.call(document.querySelectorAll('[integrity]'), (x) => {
-		return {
-			"integrity": x.getAttribute('integrity'),
-			"src": x.getAttribute(src),
-			"tagname": x.tagName.toLowerCase()
-		}
-	})
+  "iframe-allow-sandbox": (() => {
+    const link = document.createElement('a');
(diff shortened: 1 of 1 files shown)


custom_metrics/security.js
```diff
@@ -0,0 +1,21 @@
+return JSON.stringify({
+	"iframe-allow": [].map.call(document.querySelectorAll('iframe[allow]'), (x) => {
```
rviscomi, line 2: For consistency with other metrics could we use the `Array.from` method?
```suggestion
	"iframe-allow": Array.from(document.querySelectorAll('iframe[allow]')).map(iframe => {
```

rviscomi, line 2: Could you convert tabs to 2 spaces?

custom_metrics/security.js
```diff
@@ -0,0 +1,21 @@
...
+		return {
+			"allow": x.getAttribute('allow'),
+			"src": x.getAttribute(src)
```
rviscomi, line 5: Do you actually need the full `src` value? Not sure how useful that will be. If not needed you could return an array of `allow` values instead.

tomvangoethem: I wanted a way to uniquely identify `<iframe>` elements such that - if need-be - we could determine which `<iframe>`s had both `allow` and `sandbox` attribute. I think we should at least get the host of `src`.

I could change code so that we get both `allow` and `sandbox` attributes in one metric, and just get the host of the src, would that be better?

rviscomi: In that case maybe consider combining them into something like an array of attributes per element.

```js
"iframe-allow-sandbox": Array.from(document.querySelectorAll('iframe[sandbox], iframe[allow]')).map(iframe => {
  return {
    "sandbox": iframe.getAttribute('sandbox'),
    "allow": iframe.getAttribute('allow')
  }
}),
```

tomvangoethem: I added both `sandbox` and `allow` in one metric. I also added the origin, as I suspect that many iframes that have a `sandbox` attribute were added by third-party code, e.g. Google Ads (which paints a different picture than if web developers added this attribute themselves)

PR #175, opened 2020-07-27
Add CSS-in-JS libraries detection custom metrics

dooman87:
Custom metrics to find if the site is using styled-components or radium CSS-in-JS libraries.

@rviscomi Do I need to include css.js anywhere?

@LeaVerou I'll start ball rolling and let's see if we can get anywhere before deadline.

### DIFF
--- a/custom_metrics/css.js
+++ b/custom_metrics/css.js
@@ -4,18 +4,18 @@
 return JSON.stringify({
   'css_in_js': (() => {
     const CssInJsMap = {
-      'styled-components': !!document.querySelector('style[data-styled],style[data-styled-components]'),
-      'radium': !!document.querySelector('[data-radium]'),
-      'jss': !!document.querySelector('[data-jss]'),
-      'emotion': !!document.querySelector('[data-emotion]'),
-      'goober': !!document.getElementById('_goober'),
-      'merge-styles': !!document.querySelector('[data-merge-styles]'),
-      'jsx': !!document.querySelector('style[id*="__jsx-"]'),
-      'aphrodite': !!document.querySelector('[data-aphrodite]'),
-      'fela': !!document.querySelector('[data-fela-stylesheet]'),
-      'styletron': !!document.querySelector('[data-styletron],._styletron_hydrate_'),
-      'react-native-web': !!document.querySelector('#react-native-stylesheet'),
-      'glamor': !!document.querySelector('[data-glamor]')
+      'Styled Components': !!document.querySelector('style[data-styled],style[data-styled-components]'),
+      'Radium': !!document.querySelector('[data-radium]'),
+      'React JSS': !!document.querySelector('[data-jss]'),
+      'Emotion': !!document.querySelector('[data-emotion]'),
+      'Goober': !!document.getElementById('_goober'),
+      'Merge Styles': !!document.querySelector('[data-merge-styles]'),
+      'Styled Jsx': !!document.querySelector('style[id*="__jsx-"]'),
+      'Aphrodite': !!document.querySelector('[data-aphrodite]'),
+      'Fela': !!document.querySelector('[data-fela-stylesheet]'),
+      'Styletron': !!document.querySelector('[data-styletron],._styletron_hydrate_'),
+      'React Native for Web': !!document.querySelector('#react-native-stylesheet'),
+      'Glamor': !!document.querySelector('[data-glamor]')
     }
 
     const usedLibraries = []
--- a/custom_metrics/css.js
+++ b/custom_metrics/css.js
@@ -22,7 +22,7 @@ return JSON.stringify({
         let hasJSX = false;
         document.querySelectorAll('style').
             forEach((s) => {
-                hasJSX = hasJSX || s.id.indexOf('__jsx-') == 0
+                hasJSX = hasJSX || s.id.indexOf('__jsx-') === 0
             });
 
         return hasJSX;
--- a/custom_metrics/css.js
+++ b/custom_metrics/css.js
@@ -1,30 +1,24 @@
 //[almanac-css]
 // Uncomment the previous line for testing on webpagetest.org
 
-// README! Instructions for adding a new custom metric for the Web Almanac.
-//
-// Add a new key/value pair to the return object below.
-//
-// 1. The key should be a unique identifier for the metric, if possible, like it's metric ID (eg '01.12').
-// 1a. If the value is used by multiple metrics, the key should be named according to what it's measuring (eg 'link-nodes').
-// 2. If the value requires more than one line of code, evaluate it in an IIFE, eg `(() => { ... })()`. See `link-nodes`.
-// 3. Test your change by following the instructions at https://github.com/HTTPArchive/almanac.httparchive.org/issues/33#issuecomment-502288773.
-// 4. Submit a PR to update this file.
-
 return JSON.stringify({
-    'has_cssinjs_styled_components': !!(document.querySelector('style[data-styled]').length || document.querySelectorAll('style[data-styled-components]').length),
-    'has_cssinjs_radium': !!document.querySelector('[data-radium]').length,
-    'has_cssinjs_jss': !!document.querySelector('[data-jss]').length,
-    'has_cssinjs_emotion': !!document.querySelector('[data-emotion]').length,
-    'has_cssinjs_goober': !!document.getElementById('_goober'),
-    'has_cssinjs_merge-styles': !!document.querySelector('[data-merge-styles]'),
-    'has_cssinjs_jsx': (() => {
-        let hasJSX = false;
-        document.querySelectorAll('style').
-            forEach((s) => {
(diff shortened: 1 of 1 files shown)


LeaVerou:
I like where this is going. Please note it's more performant to use `document.querySelector()` instead of `document.querySelectorAll()` when all you care about is presence.

dooman87:
Yeah, you are right. Will update the PR

rviscomi:
>@rviscomi Do I need to include css.js anywhere?

Nope, just existing in this directory is all we need.

Please remove from "Draft" mode when it's ready for review.

custom_metrics/css.js
```diff
@@ -0,0 +1,30 @@
+//[almanac-css]
```
rviscomi, line 1:
```suggestion
//[css]
```

custom_metrics/css.js
```diff
@@ -0,0 +1,30 @@
...
+// Add a new key/value pair to the return object below.
+//
+// 1. The key should be a unique identifier for the metric, if possible, like it's metric ID (eg '01.12').
+// 1a. If the value is used by multiple metrics, the key should be named according to what it's measuring (eg 'link-nodes').
+// 2. If the value requires more than one line of code, evaluate it in an IIFE, eg `(() => { ... })()`. See `link-nodes`.
+// 3. Test your change by following the instructions at https://github.com/HTTPArchive/almanac.httparchive.org/issues/33#issuecomment-502288773.
+// 4. Submit a PR to update this file.
+
+return JSON.stringify({
+    'has_cssinjs_styled_components': !!(document.querySelector('style[data-styled]').length || document.querySelectorAll('style[data-styled-components]').length),
```
rviscomi, line 15: > document.querySelector('style[data-styled]').length

`length` doesn't work on `querySelector`
```suggestion
    'has_cssinjs_styled_components': !!document.querySelector('style[data-styled], style[data-styled-components]'),
```

custom_metrics/css.js
```diff
@@ -0,0 +1,21 @@
...
+// 3. Test your change by following the instructions at https://github.com/HTTPArchive/almanac.httparchive.org/issues/33#issuecomment-502288773.
+// 4. Submit a PR to update this file.
+
+return JSON.stringify({
+    'has_cssinjs_styled_components': document.querySelectorAll('style[data-styled]').length > 0 || document.querySelectorAll('style[data-styled-components]').length > 0,
+    'has_cssinjs_radium': document.querySelectorAll('[data-radium]').length > 0,
+    'has_cssinjs_jss': document.querySelectorAll('[data-jss]').length > 0,
+    'has_cssinjs_emotion': document.querySelectorAll('[data-emotion]').length > 0,
+    'has_cssinjs_goober': !!document.getElementById('_goober')
+
```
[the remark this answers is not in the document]
dooman87: Thanks, done

custom_metrics/css.js
dooman87: Yup, was too early in the day, cheers!

custom_metrics/css.js
dooman87: Done

custom_metrics/css.js
```diff
@@ -0,0 +1,24 @@
...
+  'has_cssinjs_goober': !!document.getElementById('_goober'),
+  'has_cssinjs_merge-styles': !!document.querySelector('[data-merge-styles]'),
+  'has_cssinjs_jsx': (() => {
+    let hasJSX = false;
+    document.querySelectorAll('style').forEach((s) => {
+      hasJSX = hasJSX || s.id.indexOf('__jsx-') === 0
+    });
+
+    return hasJSX;
+  })(),
```
rviscomi, line 18: Can you use an attribute selector for this?
```suggestion
  'has_cssinjs_jsx': !!document.querySelector('style[id*="__jsx-"]'),
```

custom_metrics/css.js
```diff
@@ -0,0 +1,24 @@
...
+  'has_cssinjs_aphrodite': !!document.querySelector('[data-aphrodite]'),
+  'has_cssinjs_fela': !!document.querySelector('[data-fela-stylesheet]'),
+  'has_cssinjs_styletron': !!document.querySelector('[data-styletron]') || !!document.querySelector('._styletron_hydrate_'),
```
rviscomi, line 21: Combine?
```suggestion
  'has_cssinjs_styletron': !!document.querySelector('[data-styletron], ._styletron_hydrate_'),
```

custom_metrics/css.js
```diff
@@ -0,0 +1,24 @@
+//[css]
+// Uncomment the previous line for testing on webpagetest.org
+
+return JSON.stringify({
+  'has_cssinjs_styled_components': !!(document.querySelector('style[data-styled]') || document.querySelector('style[data-styled-components]')),
```
rviscomi, line 5: Weird, one of my comments got lost.

I wanted to ask if you could combine all of these into a single array property rather than having one property per tool. Example output:

```json
{
  "css_in_js": ["Goober", "JSS"]
}
```

This makes it easier to tell if there are _any_ CSS-in-JS tools in use by checking the length of the array rather than testing each individual property.

dooman87: Yeah, was thinking about. How all this will be represented in BigQuery? Would it be still easy to calculate popularity of each framework?

rviscomi: Yes, the query could easily sort by the most popular tool names in the array. I can show you some examples when we get there.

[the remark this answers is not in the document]
dooman87: Thanks a lot for help with selectors. I updated them all.

custom_metrics/css.js
dooman87: Thanks, done

custom_metrics/css.js
dooman87: Please, have a look at new solution. As discussed it returns an array with frameworks used. A quick one, do you think it would be better to return empty array or null if there is no libraries used?

custom_metrics/css.js
dooman87: Thanks, done!

custom_metrics/css.js
```diff
@@ -0,0 +1,30 @@
...
+  'styletron': document => !!document.querySelector('[data-styletron],._styletron_hydrate_'),
+  'react-native-web': document => !!document.querySelector('#react-native-stylesheet'),
+  'glamor': document => !!document.querySelector('[data-glamor]')
+}
+
+return JSON.stringify({
+  'css_in_js': (() => {
+    const usedLibraries = []
+    for (l in CssInJsMap) {
+      if (CssInJsMap[l](document)) {
```
LeaVerou, line 23: Is there a reason why we want these tests to be functions on `document` They seem to be running synchronously, and on the same `document` node as the rest of the JS, so why not just execute inline?

dooman87:
Do you think just use ifs inside metric function? My thinking behind was
easier to extend and less fragile given there is no tests.

custom_metrics/css.js
dooman87: Do you think just use ifs inside metric function? The thinking behind a map was easier to extend and less fragile given there is no tests.

LeaVerou: No, I mean you can still have a map, of ids to booleans, not ids to functions.

custom_metrics/css.js
rviscomi: Empty array sounds good

dooman87: It's empty array now, so resolving this one :)

custom_metrics/css.js
dooman87: Gotcha, thanks! I moved map inside the function and inlined all the checks.

custom_metrics/css.js
```diff
@@ -0,0 +1,30 @@
+//[css]
+// Uncomment the previous line for testing on webpagetest.org
+
+return JSON.stringify({
+  'css_in_js': (() => {
+    const CssInJsMap = {
+      'styled-components': !!document.querySelector('style[data-styled],style[data-styled-components]'),
```
rviscomi, line 7: Rather than IDs/keys like `'styled-components'` could we use the more "human readable" names like `'Styled Components'` to avoid the need for converting later in the pipeline?

dooman87: Started doing that and "Merge Styles" looks a bit too generic. Now, I'm thinking maybe we should use npm package names? For example "@uifabric/merge-styles" and so on?

rviscomi: Let's opt for maximal readability since these values will be shown to readers in data visualizations. I think even a generic name like "Merge Styles" is preferable to its npm package name.

dooman87: Sounds good, updated

LeaVerou:
This is purely cosmetic, but do note you can build the array in one line:

```
Object.entries(CssInJsMap).filter(a => a[1]).map(a => a[0])
```

PR #172, opened 2020-07-23
New metrics and bug fixes

obto:
I've looked through all of the metrics from 2019 and created new functions to minimize the use of regex this year.

Here are the summaries:

### Metric to Function mapping

| Metric | Function to use |
| ------ | --------------- |
| 04_04.sql | inline_svg_stats |
| 04_06.sql | images.(total_*) |
| 04_06b.sql | images.sizes_values |
| 04_08.sql | image_stats.total_pictures_with_img |
| 04_09a.sql | meta-nodes should already take care what is portable here |
| 04_09b.sql | meta-nodes should already take care what is portable here |
| 04_09c.sql | meta-nodes should already take care what is portable here |
| 04_09d.sql | meta-nodes and images.sizes_values |
| 04_19.sql | videos |
| 04_26.sql | images.images_using_loading_prop |
| 08_32.sql | Using headers |
| 08_33.sql | Using headers |
| 08_35-37.sql | Using headers |
| 08_39.sql | scripts and link-nodes |
| 08_42.sql | Using headers |
| 09_02.sql | nodes_using_role |
| 09_04.sql | nodes_using_role |
| 09_05.sql | nodes_using_role |
| 09_05b.sql | nodes_using_role |
| 09_06.sql | total_nodes_with_duplicate_ids |
| 09_11.sql | heading-order from Axe https://dequeuniversity.com/rules/axe/3.0/heading-order |
| 09_13.sql | headings_order |
| 09_14.sql | shortcuts_stats |
| 09_15.sql | shortcuts_stats |
| 09_16.sql | Can use input-elements |
| 09_16b.sql | Can use input-elements |
| 09_17.sql | nodes_using_role |
| 09_19a.sql | attributes_used_on_elements |
| 09_19b.sql | attributes_used_on_elements |
| 09_19c.sql | attributes_used_on_elements |
| 09_22.sql | body_node |
| 09_24.sql | attributes_used_on_elements |
| 09_32b.sql | images.alt_tag_lengths |
| 10_02.sql | html_node |
| 10_04a.sql | link-nodes |
| 10_04b.sql | link-nodes |
| 10_07b.sql | document_title |
| 10_16.sql | length_of_h1s |
| 14_02.sql | meta-nodes |
| 14_03.sql | meta-nodes |

### Metrics unable to create a function for

| Metric | Reason |
| ------ | ------ |
| 01_18.sql | Its scanning javascript |
| 01_24.sql | Its scanning javascript |
| 01_25.sql | Its scanning javascript |
[...]

PR #177, opened 2020-07-28
Custom metrics for Markup

Tiggerito:
### Testing

### Code Architecture
- [x] **try catch logging system**
I've implemented some helper functions to enable logging errors to a log property of the almanac. Using this and try catch blocks around property code will protect us from a one error all lost situation. It will also enable us to query the errors that happen so we can fix them for the next run.

### [Markup](https://github.com/HTTPArchive/almanac.httparchive.org/issues/899)
- [x] **favicon**
- [x] **rel_alternate_mobile**
- [x] **compatMode**
- [x] **noscript**
uses and a test for GTM within the noscript
- [x] **buttons**
type attribute info
- [x] **inputs**
type attribute info
- [x] **dirs**
text direction usage
- [x] **audios**
 autoplay use
- [x] **classes**
summary of class name use
- [x] **ids**
summary of id use including a duplicate count
- [x] **images**
testing the loading attribute
- [x] **iframes**
data related to src, srcset, alt, loading, width,height attributes
- [x] **amp**
testing for amp features
- [x] **data_nosnippet**
testing for use of this attribute
- [x] **obsolete_elements**
identify obsolete elements
- [x] **svgs**
count different ways svgs are implemented
- [x] **app**
trying to identify if it is an app

Progress on https://github.com/HTTPArchive/almanac.httparchive.org/issues/899


custom_metrics/markup.js
```diff
@@ -0,0 +1,647 @@
...
+
+  return error;
+}
+
+var _custom_metrics = null;
+
+try { // whole process is placed in a try/catch so we can log uncaught errors
+
+  _custom_metrics = {
+    'doctype': document.doctype?.name ?? null,
```
rviscomi, line 55: This custom metric already exists at https://github.com/HTTPArchive/legacy.httparchive.org/blob/master/custom_metrics/doctype.js
```suggestion
```

custom_metrics/markup.js
```diff
@@ -0,0 +1,647 @@
...
+    'favicon': !!document.querySelector('link[rel*="icon"]'),
+    'viewport': document.querySelector('meta[name="viewport"]')?.getAttribute('content') ?? null,
+    'rel_alternate_mobile': !!document.querySelector('link[rel="alternate"][media][href]'),
+    'compatMode': document.compatMode,
+
+    // Get the html lang attribute if present. Was previously done via SQL which only captured the first two characts of the value (country code)
+    // Used by: SEO
+    'html_lang': document.querySelector('html')?.getAttribute('lang')?.toLowerCase(),
+
```
rviscomi, line 64: This is covered by the `html_node` metric in #172 
```suggestion
```

custom_metrics/markup.js
```diff
@@ -0,0 +1,647 @@
...
+    // Used by Markup
+    'app': (() => {   
+      try {   
+        let result = {};
+
+        //  `<div id="app">` 
+        result.app_id_present = !!document.getElementById("app");
+
+        // `<meta name="theme-color">` 
+        result.meta_theme_color_present = !!document.querySelector('meta[name="theme-color"]');
```
rviscomi, line 622: May be interesting to return the actual theme color value or null if not present.

custom_metrics/markup.js
```diff
@@ -0,0 +1,647 @@
...
+          });
+
+        result.total = nodes.length;
+
+        return result;
+      }
+      catch(e) {
+        return logError("videos", e);
+      }
+    })(),
```
rviscomi, line 242: The `videos` metric in #172 provides the same info on video elements and more, so we can remove this one.
```suggestion
```

custom_metrics/markup.js
```diff
@@ -0,0 +1,647 @@
...
+          else 
+            result.inline++;
+        });
+
+        return result;
+      }
+      catch(e) {
+        return logError("scripts", e);
+      }
+    })(),
```
rviscomi, line 216: Covered by the `scripts` metric in #172 
```suggestion
```

custom_metrics/markup.js
```diff
@@ -0,0 +1,647 @@
...
+
+    // markup info 
+    // Used by: Markup
+    'deprecated_elements': (() => {  
+      try {   
+        // https://html.spec.whatwg.org/multipage/obsolete.html#non-conforming-features
+        // Array.from(document.querySelectorAll('dfn[data-dfn-type] code')).map(e => e.innerText).join(',')
+        let result = {};
+        let deprecatedNodes = [...document.querySelectorAll('applet,acronym,bgsound,dir,noframes,isindex,keygen,listing,menuitem,nextid,noembed,plaintext,rb,rtc,strike,xmp,basefont,big,blink,center,font,multicol,nobr,spacer,tt,frameset,frame')];
```
rviscomi, line 574: See https://discuss.httparchive.org/t/use-of-deprecated-html-features-on-the-web/1416/11?u=rviscomi. There are some additional deprecated elements that I was checking for (like `marquee`) that are not present here. Could you double check to make sure this is a complete list?

custom_metrics/markup.js
```diff
@@ -0,0 +1,647 @@
...
+}
+
+var _custom_metrics = null;
+
+try { // whole process is placed in a try/catch so we can log uncaught errors
+
+  _custom_metrics = {
+    'doctype': document.doctype?.name ?? null,
+    'favicon': !!document.querySelector('link[rel*="icon"]'),
+    'viewport': document.querySelector('meta[name="viewport"]')?.getAttribute('content') ?? null,
```
rviscomi, line 57: I think this is a duplicate of the existing [`meta_viewport.js`](https://github.com/HTTPArchive/legacy.httparchive.org/blob/master/custom_metrics/meta_viewport.js) custom metric. Could you check and remove this if so?

custom_metrics/markup.js
Tiggerito: removed

custom_metrics/markup.js
Tiggerito: removed.

custom_metrics/markup.js
Tiggerito: This was a specific request from Markup. It's one way they determine if the page is an app. I'll update it to return null or the color.

custom_metrics/markup.js
Tiggerito: ok

custom_metrics/markup.js
Tiggerito: Looks like the selector does not capture them all. I'll ask the Markup team.

custom_metrics/markup.js
Tiggerito: removing

custom_metrics/markup.js
Tiggerito: The answer was work it out yourself. Only a few were missing. Done

custom_metrics/markup.js
```diff
@@ -0,0 +1,580 @@
...
+    })(),
+
+    // markup info 
+    // Used by: Markup
+    'obsolete_elements': (() => {  
+      try {   
+        // https://html.spec.whatwg.org/multipage/obsolete.html#non-conforming-features
+        // Array.from(document.querySelectorAll('dfn[data-dfn-type] code')).map(e => e.innerText).join(',')
+        let result = {};
+        let obsoleteNodes = [...document.querySelectorAll('applet,acronym,bgsound,dir,frame,frameset,noframes,isindex,keygen,listing,menuitem,nextid,noembed,plaintext,rb,rtc,strike,xmp,basefont,big,blink,center,font,marquee,multicol,nobr,spacer,tt')];
```
obto, line 507: Please break this line up into multiple lines
```suggestion
        let obsoleteNodes = [...document.querySelectorAll(
          'applet,acronym,bgsound,dir,frame,frameset,noframes,isindex,keygen,listing,menuitem,nextid,noembed,' +
          'plaintext,rb,rtc,strike,xmp,basefont,big,blink,center,font,marquee,multicol,nobr,spacer,tt'
        )];
```

custom_metrics/markup.js
```diff
@@ -0,0 +1,580 @@
...
+}
+catch(e) { // probably an un caught exception
+  logError("general", "Failed to create the custom metrics object", e);
+  _custom_metrics = {};
+}
+
+if (!_custom_metrics) { // should not be possible
+  _custom_metrics = {};
+  logError("general", "custom metrics object was missing");
+}
```
obto, line 573: If it's not possible, why are we including it?

custom_metrics/markup.js
obto: Bonus points if you alphabetize the list, so its easier to see what is and is not included

custom_metrics/markup.js
```diff
@@ -0,0 +1,580 @@
...
+        nodes.forEach((n) => {
+          if (n.innerHTML.match(/googletagmanager\.com/gi)) 
+            result.iframe_googletagmanager_count++;
+        });
+
+        return result;
+      }
+      catch(e) {
+        return logError("noscripts", e);
+      }
```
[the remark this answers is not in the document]
Tiggerito: Looks neater. Do we have time to do refactoring like this?

custom_metrics/markup.js
```diff
@@ -0,0 +1,580 @@
...
+          });
+
+        result.total = nodes.length;
+
+        return result;
+      }
+      catch(e) {
+        return logError("audios", e);
+      }
+    })(),
```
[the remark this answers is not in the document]
Tiggerito: I also had video with the same pattern, but that was removed due to duplication. Would it then make sense to have almanac.js do audio and video?

I think inputs is also more broadly covered by almanac.js

If I'm going to have to parse the node list in SQL for video, I might as well use the same pattern for audio and inputs?

So I think remove audio an inputs if you add audio to almanac.js just like video?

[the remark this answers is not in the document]
obto: Keep the properties in here so we can get this merged first. We can worry about further deduplication and organization later. Lets make sure we get all the metrics locked in.

If these are the only two sections using this, then dont bother changing them. We can bring in this code later on in the main almanac.js file perhaps

custom_metrics/markup.js
Tiggerito: > If it's not possible, why are we including it?

Because I'm careful. I've done a lot of coding in my time and seen code hit that I didn't think was possible. 

Like I do try catches inside the logError. JavaScript exception objects are horrible.

I can remove it. In this case the result would most likely be the custom metric is blank instead of the message.

custom_metrics/markup.js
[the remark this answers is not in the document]
obto: That's up to you. Can you can refactor the code and confirm its all working in the next ~16 hours? Or would that be a bit tight?

custom_metrics/markup.js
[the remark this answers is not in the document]
Tiggerito: How about I add a comment to remind us that they could be de-duped or optimised?

custom_metrics/markup.js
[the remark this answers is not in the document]
Tiggerito: I suspect no one else will use it this year. I say we stick with what has been tested and later tidy it up. I guess when we have a two week deadline next year 😲

Does this get run every month, or just the one month a year?  Would we be able to clean things up over time?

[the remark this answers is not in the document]
rviscomi: I assumed having try/catch in each function meant that any exceptions would only affect the one metric and not spoil the entire batch.

Tiggerito:
Yes. if a property errors it's content is an error message and that message is added to a logs property. So we can query to see if and what errors happened.

Tiggerito:
From what I've seen, a custom metrics will blank out if it has an uncaught exception. So it looks like one can't kill the whole batch. 

I'm just protecting at two other levels so that one property can't kill the whole custom metric, and so we have data to fix things with later.

custom_metrics/markup.js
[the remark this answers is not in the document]
Tiggerito: referenced this suggestion in the code. We can refactor later

custom_metrics/markup.js
Tiggerito: Removing it.

Tiggerito:
I couldn't sleep (only 2 hours last night) and ended up thinking about the improvements @obto suggested for my error handling. Definitely not for this release, but I came up with a further possible refinement. It could even be cool to build in async/Promise ability 🤪

```
let _cm = {}; // what we will return.

// generic function to add events to the log
function logEvent(type, category, messageOrException, exception = null) {
    let event = {type: type, category: category}; 

    // ... create the event ...

    if (!_cm.log) _cm.log = [];
    _cm.log.push(event);
    return event;
}

// helper to safely set property values
function setProperty(name, functionOrData) {
    if (_cm[name]) {
        logEvent("error", name, "already in use");
    } else {
        try {
            if(functionOrData instanceof Function) 
                _cm[name] = functionOrData(_cm, name);
            else 
                _cm[name] = functionOrData;
        } catch (e) {
            _cm[name] = logEvent("error", name, e);
        }
    }
    return _cm[name];
}

try {
    // add custom code here

    // example of adding a simple property
    setProperty("test", {hello: "world"});

    // example of adding a property using a function
    setProperty("test2", (cm, name) {
        logEvent("info", name, "Made it to test2");

        return {another_test: cm.test.hello};
    });

} catch (e) {
    logEvent("error", "global", e);
}

return _cm; // the final return
```

PR #178, opened 2020-07-28
Custom metrics for SEO robots.txt

Tiggerito:
A separate custom metrics (robots_txt.js) that fetches the robots.txt file and gathers some basic info.

### Testing

Progress on https://github.com/HTTPArchive/almanac.httparchive.org/issues/908


custom_metrics/robots_txt.js
```diff
@@ -0,0 +1,39 @@
+//[robots_txt]
+return fetch('/robots.txt')
+  .then(r => {
+    let result = {};
+    result.redirected = !!r.redirected;
+    result.status = r.status;  
+    return r.text().then(t => {
+      result.size = t.length;
+      result.comment_lines = t.match(/^\s*#\s*(.*)$/gm)?.length;
+     // result.empty_lines = t.match(/^\s*$/gm)?.length;
```
rviscomi, line 10:
```suggestion
```

custom_metrics/robots_txt.js
```diff
@@ -0,0 +1,39 @@
...
+      result.allow_lines = t.match(/^\s*allow\s*:\s*(.*?)\s*$/gmi)?.length;
+      result.disallow_lines = t.match(/^\s*disallow\s*:\s*(.*?)\s*$/gmi)?.length;
```
rviscomi, line 12: In [this test](https://www.webpagetest.org/custom_metrics.php?test=200728_C6_5dbc6b7cfb804d6b7c406a88832312b8&run=1&cached=0) `disallow_lines` was omitted. Is that intentional or is it expected to have `disallow_lines: 0` in the response?

Similarly for `comment_lines`.

Tiggerito: Something I think my code may have mixed results for. Is it better to return a default value or assume it if the property is not present? 

In most cases I think I return 0. In some I did not to reduce the bloat on the data. Like the meta robots section that has a lot of properties that mostly return their default value, so I exclude them unless set. 

I'll be very defensive in all my SQL to make sure it deals with missing properties, nulls etc.

rviscomi: I think having consistency in the results would be beneficial and reduce/eliminate the need for defensiveness in the SQL. Considering that these metrics can be queried by anyone, it'd be good for them not to need to know implementation details in order to query them effectively.

Tiggerito: Fixed.

PR #179, opened 2020-07-29
Privacy custom metrics: IAB Transparency and Consent Framework

max-ostapenko:
- [x] iab_tcf function
Returns `1` if IAB TCF is integrated on the page.

Progress on https://github.com/HTTPArchive/almanac.httparchive.org/issues/913

### DIFF
--- a/custom_metrics/privacy.js
+++ b/custom_metrics/privacy.js
@@ -3,6 +3,7 @@
 
 return JSON.stringify({
   privacy_wording_links: (() => {
+    // Number of privacy related links are available on the page, eg. 'Privacy policy'.
     // Wording sourced from: https://github.com/RUB-SysSec/we-value-your-privacy/blob/master/privacy_wording.json
     //words = privacy_wording.map(country => country.words).filter((v, i, a) => a.indexOf(v) === i).flat().sort().join('|');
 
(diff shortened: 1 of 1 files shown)


max-ostapenko:
@ydimova looks good to you?
@rviscomi need to add one more for privacy chapter

custom_metrics/privacy.js
```diff
@@ -14,4 +15,10 @@ return JSON.stringify({
       a.innerText.match(pattern)
```
rviscomi, line 15: By the way, do you care to know which privacy text matched, or is it sufficient to know that any one of them matched?

max-ostapenko: We are sufficient with just finding a proper threshold on the number of matched links, probably looking at >0.
Obviously there may be some legal topic websites that have plenty of links, and some localised titles that were missed here.

I was actually thinking to just add a.innerText to wpt_bodies.anchors and run matching in BQ, but trying to keep data small.
Do you think it makes more sense because links information can be reused for other cases?

max-ostapenko:
@ydimova is it important which version of TSF is implemented?

custom_metrics/privacy.js
rviscomi: I wouldn't recommend dumping all anchor text, but if you want to return the matches from this narrow list of interest that makes sense to me so you can graph the popularity of different privacy phrases.

I'll merge this PR as-is so we can get something in before the cutoff, but please submit a PR ASAP if you have any additional enhancements. I'll be syncing the test server with the custom metrics tomorrow.

max-ostapenko:
Yeah, it was good enough for a first year.
Thanks for merging.

PR #180, opened 2020-07-29
Add Sass custom metrics

LeaVerou:
Details here: https://github.com/leaverou/css-almanac/issues/39#issuecomment-665714420

### DIFF
--- a/custom_metrics/sass.js
+++ b/custom_metrics/sass.js
@@ -1,0 +1,309 @@
+//[sass]
+const SassFunctions = [
+	// Color
+	"color.adjust", "adjust-color",
+	"color.scale", "scale-color",
+	"color.change",
+	"adjust-hue",
+	"opacity",
+	"darken", "lighten",
+	"desaturate", "saturate",
+	"opacify", "transparentize",
+	"fade-in", "fade-out",
+
+	...[
+		"red", "green", "blue",
+		"hue", "saturation", "lightness",
+		"alpha", "complement",
+		"grayscale",
+		"ie-hex-str",
+		"invert",
+		"mix"
+	].flatMap(n => ["color." + n, n]),
+
+	// List
+	"list.separator", "list-separator",
+
+	...[
+		"append", "index", "is-bracketed", "join",
+	 "length", "set-nth", "nth", "zip"
+	].flatMap(n => ["list." + n, n]),
+
+	// Map
+	...[
+		"get", "has-key", "keys", "merge", "remove", "values"
+	].flatMap(n => ["map." + n, "map-" + n]),
+
+	// Math
+	...[
+		"ceil", "floor", "round",
+		"clamp", "min", "max",
+		"abs", "comparable", "unit", "percentage", "random"
+	].flatMap(n => ["math." + n, n]),
+	"math.hypot", "math.log", "math.pow", "math.sqrt",
+	"math.sin", "math.cos", "math.tan",
+	"math.asin", "math.acos", "math.atan", "math.atan2",
+	"math.is-unitless", "unitless",
+
+	// Meta
+	"meta.load-css",
+	"meta.module-functions", "meta.module-variables",
+	...[
+		"call", "get-function", "keywords",
+		"content-exists", "feature-exists", "function-exists", "global-variable-exists", "variable-exists", "mixin-exists",
+		"inspect", "type-of"
+	].flatMap(n => ["meta." + n, n]),
+
+	"if"
+];
+
+const sortObject = obj => Object.fromEntries(Object.entries(obj).sort((a, b) => b[1] - a[1]));
+
+function analyzeSCSS(scss, ret) {
+	if (!scss) {
+		return null;
+	}
+
+	// Drop (most) comments to minimize false positives since we have to parse with regexes
+	// We want to be conservative here and err on the side of NOT removing comments
+	scss = scss.replace(/^\s*\/\*[\s\S]*?\*\//mg, "");
+	scss = scss.replace(/^\s*\/\/.+/mg, "");
+
+	// Most popular variable names
+	ret.variables = {};
+	scss.replace(/\$[\w-]+\b/g, variable => {
+		ret.variables[variable] = (ret.variables[variable] || 0) + 1;
+	});
+	ret.variables = sortObject(ret.variables);
+
+	// Mixins
+	// Note: We are not taking the Indented Mixin Syntax into account
+	ret.mixins = {};
+	scss.replace(/@mixin\s+([\w-]+)(?:\((.+?)\))?/g, ($0, name, args) => {
+		// Note: - and _ are equivalent in mixin names and refer to the same mixin
+		//         we are not taking that into account here. This applies to functions too (below)
+		ret.mixins[name] = {
+			// Default values can't have commas, so split should work
+			args: args?.split(/\s*,\s*/), // TODO separate name from default value?
+			calls: scss.match(RegExp("@include\\s+" + name + "\\b", "gi"))?.length
+		};
+	});
+
+	// Custom functions
+	ret.functions = {};
+	let lastName; // used to track down the nearest @return after a @function
+	scss.replace(/@function\s+([\w-]+)(?:\((.+?)\))?|@return (.+)/g, ($0, name, args, returnValue) => {
+		if ($0.indexOf("@return") === 0 && lastName) {
+			ret.functions[lastName].return = returnValue;
+			lastName = undefined;
+			return $0;
+		}
+
+		lastName = name;
+		ret.functions[name] = {
+			// Default values can't have commas, so split should work
+			args: args?.split(/\s*,\s*/), // TODO separate name from default value?
+			calls: scss.match(RegExp("\\b" + name + "\\(", "gi"))?.length
+		};
+	});
+
+	// TODO Measure usage of Sass functions
+	ret.functionCalls = {};
+	let functionCallRegex = new RegExp(`(?<![-.\\w])(${SassFunctions.join("|").replace(".", "\\.")})\\(`, "gi");
+	scss.replace(functionCallRegex, ($0, name) => {
+		ret.functionCalls[name] = (ret.functionCalls[name] || 0) + 1;
+	});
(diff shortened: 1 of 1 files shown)


LeaVerou:
@mirisuzanne I would appreciate your input on the heuristics as well, since you're far more experienced with Sass than I am.

custom_metrics/sass.js
```diff
@@ -0,0 +1,309 @@
...
+if (typeof $WPT_BODIES !== "undefined") {
+	// in WPT, we already have the bodies
+	stylesheets = $WPT_BODIES.filter(request => request.type == "Stylesheet")
+		.map(file => {
+			return {url: file.url, body: file.response_body};
+		});
+}
+else {
+	// Outside WPT, use fetch()
+	stylesheets = await Promise.all($$("link[rel=stylesheet]").map(async link => {
```
rviscomi, line 205: This code will always run in WPT. Could we remove this `else`? Alternatively, is it feasible to get what we need using qSA and fetch?

custom_metrics/sass.js
```diff
@@ -0,0 +1,309 @@
...
+
+results.sourcemaps = {
+	count: sourcemapURLs.length,
+	ext: {}
+};
+
+if (sourcemapURLs.length === 0) {
+	return results;
+}
+// console.log("sourcemap URLs", sourcemapURLs);
```
rviscomi, line 248:
```suggestion
```

custom_metrics/sass.js
```diff
@@ -0,0 +1,309 @@
...
+	let scss = 0;
+
+	sources = sources.map(s => {
+		let url = new URL(s, base);
+		let ext = url.pathname.match(/\.(\w+)$/)?.[1];
+
+		if (ext) {
+			results.sourcemaps.ext[ext] = (results.sourcemaps.ext[ext] || 0) + 1;
+
+			if (ext === "scss") {
```
rviscomi, line 271: Naive question, are there any other signals that the source is Sass? Is `.sass` a thing?

custom_metrics/sass.js
```diff
@@ -0,0 +1,309 @@
...
+
+	if (scss > 0) {
+		if (json.sourcesContent) {
+			// Source is already here, no more requests needed, yay!
+			return json.sourcesContent;
+		}
+
+		let code = await Promise.all(sources.map(async s => {
+			let response = await fetch(s);
+			let text = response.ok? await response.text() : "";
```
rviscomi, line 287: 🤯

custom_metrics/sass.js
```diff
@@ -0,0 +1,309 @@
...
+	count: sourcemapURLs.length,
+	ext: {}
+};
+
+if (sourcemapURLs.length === 0) {
+	return results;
+}
+// console.log("sourcemap URLs", sourcemapURLs);
+
+// Assumption: Either all sources are SCSS or none.
```
rviscomi, line 250: What does this imply? Will it fail if one source is not SCSS?

custom_metrics/sass.js
LeaVerou: Yes, it is, just different syntax with less punctuation AFAIK (never used it). Not sure of the relative popularity between the two. I can try to tackle that as well tonight if you want. Some of the metrics will be the same, some of the regexes will be slightly different.

custom_metrics/sass.js
LeaVerou: ??

custom_metrics/sass.js
LeaVerou: No. If you look at the sample output, many of the sources are a mix of CSS and SCSS and it does fine.
Basically what it means is that we read & join the files if at least one is SCSS.

custom_metrics/sass.js
LeaVerou: Sure, if code size matters. I have it there because this way it's easy to test the code with a simple `import()` in the console, without the whole wait time of WPT. But no problem removing it!

custom_metrics/sass.js
```diff
@@ -0,0 +1,308 @@
...
+	// @extend
+	ret.extends = {};
+	scss.replace(/@extend\s*(.+)(?=\s*(?:[;}]|$))/g, ($0, selector) => {
+		ret.extends[selector] = (ret.extends[selector] || 0) + 1;
+	});
+
+	// @error
+	ret.errors = scss.match(/@error (.+)/g)?.map(e => e.slice(7));
+
+	// CSS variables that are set with Sass variables
```
mirisuzanne, line 167: This may be over-specific, or need additional tests:

- custom props defined using other sass: `--<var>: #{<anything>};` (We often use functions here rather than variables)
- custom props **named by** sass: `--#{<var>}: <anything>;`

Might be interesting to look for the `#{}` interpolation pattern other places as well. It can be used for defining selectors, property names, etc – often inside mixins or loops.

custom_metrics/sass.js
rviscomi: That'd be great, thanks.

custom_metrics/sass.js
rviscomi: lol nevermind, I assumed the `?` was some new usage of nullish coalescing but upon closer inspection it's just a ternary operator (it was late 😅)

custom_metrics/sass.js
rviscomi: Ok, I didn't realize it was a different flavor. Limiting the scope of this SGTM for simplicity and ensuring that we can merge ASAP.

So to be clear the PR title refers to "Sass" but the corresponding file extension is `.scss` and `.sass` is something else?

mirisuzanne: The overall language is Sass. It has a mostly-legacy whitespace syntax (`.sass`) and a more popular CSS-superset syntax (`.scss`). The "Sass" name does refer to both syntax/file-extenions.

custom_metrics/sass.js
```diff
@@ -0,0 +1,308 @@
...
+	scss.replace(/\$[\w-]+\b/g, variable => {
+		ret.variables[variable] = (ret.variables[variable] || 0) + 1;
+	});
+	ret.variables = sortObject(ret.variables);
+
+	// Mixins
+	// Note: We are not taking the Indented Mixin Syntax into account
+	ret.mixins = {};
+	scss.replace(/@mixin\s+([\w-]+)(?:\((.+?)\))?/g, ($0, name, args) => {
+		// Note: - and _ are equivalent in mixin names and refer to the same mixin
```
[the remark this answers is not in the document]
LeaVerou: It is possible that if that is utilized we will miss calls for mixins/custom functions, but I'm ok with a few false negatives than with replacing underscores with hyphens across the board, which seems very fragile without a parser.

custom_metrics/sass.js
```diff
@@ -0,0 +1,308 @@
...
+		ret.mixins[name] = {
+			// Default values can't have commas, so split should work
+			args: args?.split(/\s*,\s*/), // TODO separate name from default value?
+			calls: scss.match(RegExp("@include\\s+" + name + "\\b", "gi"))?.length
+		};
+	});
+
+	// Custom functions
+	ret.functions = {};
+	let lastName; // used to track down the nearest @return after a @function
```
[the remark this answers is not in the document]
LeaVerou: I'm not sure either, but I thought it might give us insight on what the function does. Good point about multiple returns, I should capture all of them.

custom_metrics/sass.js
LeaVerou: > As far as I can tell, the .sass syntax is far less popular, but does have a dedicated niche following. It would be interesting to know how much of a following that adds up to…

The code does count all extensions, so it will still give us info on the relative popularity of scss vs sass, it just won't analyze the sass files. 

> Are we capturing and comparing file/path names to compare how sass files are structured? It would be interesting to know if there are common file/directory naming structures, or how many people are using Sass loaded from npm modules…

Any ideas what we can store for this that is sufficiently compact?

> This may be over-specific, or need additional tests:

Oh good point, I even saw this in the Bootstrap SCSS files but wasn't sure it's common enough to study.
Selectors are *really* hard to track down without a parser I'm afraid :(

custom_metrics/sass.js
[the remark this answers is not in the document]
LeaVerou: Alright, it now captures multiple returns. I updated the [gist with the sample data](https://gist.github.com/LeaVerou/5d02efb07945d595b70f5ecdc69057b7) too.
From what I see there is no other `@rule` that takes an `@return`, so it's safe to assume that any `@return` we find after an `@function` is for that function?

custom_metrics/sass.js
LeaVerou: Alright, I just pushed a heuristic to measure CSS variables whose name is set via Sass interpolation.

LeaVerou:
Alright, I think I've made all changes requested. :)

custom_metrics/sass.js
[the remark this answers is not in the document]
mirisuzanne: That sounds right to me, I can't think of a case that would break that rule.

custom_metrics/sass.js
```diff
@@ -0,0 +1,295 @@
...
+		ret.extends[selector] = (ret.extends[selector] || 0) + 1;
+	});
+
+	// @error
+	ret.errors = scss.match(/@error (.+)/g)?.map(e => e.slice(7));
+
+	// CSS variables that are set with Sass variables
+	// Note that this will fail on multiline values (it will return the first line only)
+	ret.variablesCombined = {value: {}, name: {}};
+	scss.replace(/(?<=^|\s)--([\w-]+):\s*(.*#\{.*\$.+\}.*)\s*$/gm, ($0, name, value) => {
```
mirisuzanne, line 170:
```suggestion
	scss.replace(/(?<=^|\s)--([\w-]+):\s*(.*#\{.+\}.*)\s*$/gm, ($0, name, value) => {
```

Suggest not requiring the `$` to trigger this, just capture any time `#{}` is used for Sass-defined custom props.

LeaVerou: Done!

LeaVerou:
Alright. I also just pushed another change to sort mixins and functions by calls in descending order. It should be good to go.

PR #184, opened 2020-07-30
A11Y custom metrics

obto:
Added metrics to detect the following:

| Metric in a11y prop | Used for |
| ------------------- | -------- |
| tables | Tables with caption elements |
| file_extension_alts | Alt text that ends in an image extension |
| title_and_alt | How often is the title attribute with the alt attribute in the same element |
| no_destination_anchors | \<a\> elements with an href value of: nothing, #, or javascript:void |
| th_with_scope_attribute | How many \<th\> elements specify the scope attribute (col or row) |
| tables.total_with_presentational | how many tables are marked as a presentational |
| total_anchors_with_role_button | \<a\> elements with role=”button” |
| total_role_tab_with_selected_and_controls | Elements with role="tab", aria-selected and aria-controls attributes |
| placeholder_but_no_label | Form controls with a non-empty placeholder attribute but no label element, aria-labelledby or aria-labelelement |
| divs_or_spans_as_button_or_link | divs/spans with a role=”button” or role=”link” |
| does_page_use_sr_only_classes | Does a page use the css class sr-only or visually-hidden |
| td_with_headers_attribute | how many \<td\> elements have a headers attribute |

Still in progress

Progress on https://github.com/HTTPArchive/almanac.httparchive.org/issues/907


custom_metrics/almanac.js
```diff
@@ -694,4 +694,190 @@ return JSON.stringify({
   'length_of_h1s': (() => {
     return [...document.querySelectorAll('h1')].map(node => node.innerText.length);
   })(),
+
+  'a11y': (() => {
```
rviscomi, line 698: Can you create a new `accessibility.js` file for these metrics? At a glance it doesn't look like they depend on any of the "global" functions within `almanac.js`.

obto:
@rviscomi This is now ready for review. I have unit tested each metric to be working correctly. I'll start running the script on WPT now

obto:
Here are the results from WPT. All green lights are go.

https://docs.google.com/spreadsheets/d/1yzJJYZYkGuQDLaruj5cd18VSuhYHExk5u94KkSB8y28/edit?usp=sharing

custom_metrics/a11y.js
```diff
@@ -0,0 +1,243 @@
...
+    }
+
+    return {
+      total_alt: document.querySelectorAll('*[alt]').length,
+      total_title: document.querySelectorAll('*[title]').length,
+      total_both: has_both_title_and_alt.length,
+      total_alt_same_as_title,
+    };
+  }),
+  no_destination_anchors: captureAndLogError(() => {
```
rviscomi, line 134: Is there anything this function does that isn't already covered by https://github.com/HTTPArchive/legacy.httparchive.org/blob/master/custom_metrics/wpt_bodies.js#L144 and https://github.com/HTTPArchive/legacy.httparchive.org/blob/master/custom_metrics/almanac.js#L283?

Could we remove if duplicate?

custom_metrics/a11y.js
```diff
@@ -0,0 +1,243 @@
...
+      }
+
+      // Explicit label
+      const id = (element.getAttribute('id') || '').trim;
+      if (id.length > 0) {
+        const element = document.getElementsByName(id);
+        if (element) {
+          return true;
+        }
+      }
```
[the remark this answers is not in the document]
obto: Great catch. This is now fixed

custom_metrics/a11y.js
obto: They are detecting empty links as same page links... which is honestly probably better than what I'm doing here.

Deleting this as its a dupe.

PR #186, opened 2020-07-31
[sass] Add constants usage count (e & pi)

LeaVerou:
This counts how many times (if any) the constants for e and π are used.

Progress on https://github.com/HTTPArchive/almanac.httparchive.org/issues/898

### DIFF
--- a/custom_metrics/sass.js
+++ b/custom_metrics/sass.js
@@ -189,6 +189,11 @@ function analyzeSCSS(scss, ret) {
 		"class": scss.match(/&\.\w+/g)?.length,
 		"attr": scss.match(/&\[\w+/g)?.length,
 	};
+
+	ret.constants = {
+		e: scss.match(/\bmath\.$e\b/g)?.length,
+		pi: scss.match(/\bmath\.$pi\b/g)?.length
+	};
 }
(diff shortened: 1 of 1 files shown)


custom_metrics/sass.js
```diff
@@ -189,6 +189,11 @@ function analyzeSCSS(scss, ret) {
 		"class": scss.match(/&\.\w+/g)?.length,
 		"attr": scss.match(/&\[\w+/g)?.length,
 	};
+
+	ret.constants = {
+		e: scss.match(/\bmath\.$e\b/g)?.length,
```
rviscomi, line 194: Does the `$` need to be escaped?

LeaVerou: Oops, yes! Oh well, I suppose no point fixing it now.

LeaVerou:
> Are we also able to measure these values in the CSS output using Rework? Not necessarily the SASS constants, but the values themselves.

No because they are mostly used in calculations. Anyhow, not that important.

rviscomi:
> No because they are mostly used in calculations. Anyhow, not that important.

😞 ok, closing this PR

Issue #188, opened 2020-09-05
wpt_bodies meta description and robots gathering is invalid as the selector used is case sensitive


Tiggerito:
https://github.com/HTTPArchive/legacy.httparchive.org/blob/master/custom_metrics/wpt_bodies.js

It seems that name attribute values are case sensitive, so this line will only get a subset of meta descriptions.

`target.total = [...d.querySelectorAll('head meta[name="description"]')].map(e => {`

The line should be:

`target.total = [...d.querySelectorAll('head meta[name="description"i]')].map(e => {`

Same goes for the selectors in the calculateAllRobots function.

rviscomi:
Fixed by https://github.com/HTTPArchive/legacy.httparchive.org/pull/189

PR #189, opened 2020-09-06
Added i to make selectors case insensitive

Tiggerito:
#188 

I checked all the custom metrics for potential issues with them using css selectors that needed to be set to case insensitive. This pull adds i to them to make them case insensitive.

I also found potential issue with third-parties.js but it's code is auto generated. It uses selectors to detect the generator meta tag, but I think will only capture that if generator is lowercase.

I did some basic testing to check a a few cases from each file to see that it was now being case insensitive, and that the scripts ran without errors.

### DIFF
--- a/custom_metrics/almanac.js
+++ b/custom_metrics/almanac.js
@@ -481,9 +481,9 @@ return JSON.stringify({
     // Gets metadata about the AMP plugin, if present.
     // Used by 14.2, 14.3, 14.4.
     try {
-      var metadata = document.querySelector('meta[name=generator][content^="AMP Plugin"]');
+      var metadata = document.querySelector("meta[name='generator'i][content^='AMP Plugin'i]");
       if (metadata) {
-        return metadata.getAttribute('content');
+        return metadata.getAttribute('content'); // this can now vary by case
       }
     } catch (e) {
       return null;
--- a/custom_metrics/almanac.js
+++ b/custom_metrics/almanac.js
@@ -483,7 +483,7 @@ return JSON.stringify({
     try {
       var metadata = document.querySelector("meta[name='generator'i][content^='AMP Plugin'i]");
       if (metadata) {
-        return metadata.getAttribute('content'); // this can now vary by case
+        return metadata.getAttribute('content');
       }
     } catch (e) {
       return null;
--- a/custom_metrics/markup.js
+++ b/custom_metrics/markup.js
@@ -53,8 +53,8 @@ var _custom_metrics = null;
 try { // whole process is placed in a try/catch so we can log uncaught errors
 
   _custom_metrics = {
-    'favicon': !!document.querySelector('link[rel*="icon"]'),
-    'rel_alternate_mobile': !!document.querySelector('link[rel="alternate"][media][href]'),
+    'favicon': !!document.querySelector('link[rel*="icon"i]'),
+    'rel_alternate_mobile': !!document.querySelector('link[rel="alternate"i][media][href]'),
     'compatMode': document.compatMode,
 
     // noscript tag use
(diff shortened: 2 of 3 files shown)


Tiggerito:
@bazzadp @rviscomi Here's my pull request and related bug to fix the case sensitive issue in the future.

custom_metrics/almanac.js
```diff
@@ -481,9 +481,9 @@ return JSON.stringify({
     // Gets metadata about the AMP plugin, if present.
     // Used by 14.2, 14.3, 14.4.
     try {
-      var metadata = document.querySelector('meta[name=generator][content^="AMP Plugin"]');
+      var metadata = document.querySelector("meta[name='generator'i][content^='AMP Plugin'i]");
       if (metadata) {
-        return metadata.getAttribute('content');
+        return metadata.getAttribute('content'); // this can now vary by case
```
bazzadp, line 486: Do we really need this?
```suggestion
        return metadata.getAttribute('content');
```

custom_metrics/markup.js
```diff
@@ -560,7 +560,7 @@ try { // whole process is placed in a try/catch so we can log uncaught errors
         result.app_id_present = !!document.getElementById("app");
 
         // `<meta name="theme-color">` 
-        result.meta_theme_color = document.querySelector('meta[name="theme-color"]')?.getAttribute('content') ?? null;
+        result.meta_theme_color = document.querySelector('meta[name="theme-color"i]')?.getAttribute('content') ?? null;
```
bazzadp, line 563: Nothing to do with this PR but this will only return the first one. Can't imagine why someone would want to specify multiple theme-colors but if they do, wouldn't the last one (not the first one) take effect?

Probably a real edge case, and can be totally ignored if you want (especially in terms of this PR), but since everything else uses querySelectorAll and I noticed this one used querySelector thought I'd ask.

bazzadp (approved):
Not my area of expertise but LGTM!

Two comments for you to consider but nothing to hold up my approval here.

custom_metrics/almanac.js
Tiggerito: I'm altering someone else's data, so thought it would be good to include a clue if they had a problem?

custom_metrics/markup.js
Tiggerito: To be honest, this is only used to test if it is present or not. I probably passed back the value in case the chapter suddenly wanted it. 

Not sure how the recipient systems would deal with multiple entries. I know with Google and SEO things can vary per tag. First title, all meta descriptions, most restrictive robots tag etc.

bazzadp: Grand. Resolving.

custom_metrics/almanac.js
bazzadp: I think that’s what Git History is for so I’d remove it. It’s all of our code/data now since it’s a public repo 😀

Tiggerito: I'm still learning this world of cloud collaboration and version control. My poor memory and frequent need to reverse engineer probably means I comment a lot more than most. And that's when I'm working on my own code.

rviscomi:
Thanks @Tiggerito! Do you have a WPT test result to show how this works? The changes LGTM by eye but it's good to verify with a test.

Tiggerito:
> Thanks @Tiggerito! Do you have a WPT test result to show how this works? The changes LGTM by eye but it's good to verify with a test.

Here's a test using my special test page:

https://www.webpagetest.org/custom_metrics.php?test=200909_7B_f451d14269f07cf45930b4cfd35172e9&run=1&cached=0

I checked a few of the main scenarios. e.g. it is now picking up the case where description is written Description.

I also checked the changes for markup, and they looked good:

https://www.webpagetest.org/custom_metrics.php?test=200909_12_e2abd832b01f75afa32921b164c3f119&run=1&cached=0

rviscomi (approved):

PR #193, opened 2020-09-30
added fetchWithTimeout and error checking for when it does time out

### LINKED ISSUE #192: Add a shorter timeout for fetches in custom metrics
An asynchronous `fetch` in a custom metric could take ~30 seconds before timing out. Rather than wait for the promise to reject, race the fetch against a shorter timeout of ~5-10 seconds and resolve the promise at the sooner of the two async events.

This would help ensure that the custom metrics don't interfere as much with the overall crawl rate, as 30 seconds * 7 million URLs * 2 runs per client (desktop, mobile) definitely adds up.

Here are the instances of `fetch` in the custom metrics: https://github.com/search?q=fetch+repo%3AHTTPArchive%2Flegacy.httparchive.org+path%3Acustom_metrics&type=Code&ref=advsearch&l=&l=

Tiggerito:
fixes #192 

I tested each with a success scenario with the 5000ms timeout and a fail scenario caused by setting the timeout to 10ms.

ecommerce and robots_txt return this on timeout:

`{"message":"The user aborted a request.","error":{}}`

sass was a bit more complex. I don't know how to do a real world test for this. 

It uses async calls. One fetch was already in a try/catch which I tested by faking the scenario that causes the fetches. It would behave in the same way any error would have been handed before, eats it.

I added a try/catch to the  second fetch so that it behaves in the same way as the response not being ok.

### DIFF
--- a/custom_metrics/ecommerce.js
+++ b/custom_metrics/ecommerce.js
@@ -8,16 +8,21 @@
 // 3. Test your change by following the instructions at https://github.com/HTTPArchive/almanac.httparchive.org/issues/33#issuecomment-502288773.
 // 4. Submit a PR to update this file.
 
+function fetchWithTimeout(url) {
+  var controller = new AbortController();
+  setTimeout(() => {controller.abort()}, 5000);
+  return fetch(url, {signal: controller.signal});
+}
 
 return Promise.all([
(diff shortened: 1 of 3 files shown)


rviscomi (approved):
Thanks for fixing these @Tiggerito!

PR #194, opened 2020-10-05
include href attribute for SRI custom metric

tomvangoethem:
`<link>` elements can have SRI, but use a `href` attribute to point to the resource

### DIFF
--- a/custom_metrics/security.js
+++ b/custom_metrics/security.js
@@ -13,7 +13,7 @@ return JSON.stringify({
   "sri-integrity": Array.from(document.querySelectorAll('[integrity]')).map((element) => {
     return {
       "integrity": element.getAttribute('integrity'),
-      "src": element.getAttribute('src'),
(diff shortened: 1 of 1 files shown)


rviscomi (approved):
LGTM. Does this mean the August data is incorrect?

tomvangoethem:
There's still the data for `<script>` elements, which have the `src` attribute (which make up the majority of elements using SRI), so it's not that much of an issue.
I queried just the script elements [here](https://github.com/HTTPArchive/almanac.httparchive.org/blob/b89d1071f5195ae900dd9b9bc13bc464603472df/sql/2020/11_Security/sri_popular_hosts.sql)

rviscomi:
Sounds good. Be aware that the October crawl has already started, so the `link[href]` data won't be in before the November crawl, which will be ready in December. So I hope you're not waiting on it for the 2020 Almanac 😁

PR #196, opened 2020-11-04
Detection using data-jss essentially detects JSS core, 

kof:
react-jss and material-ui all use jss core, it is confusing if this metric is tied to react-jss library by naming


rviscomi:
cc @LeaVerou

LeaVerou:
That was @dooman87's (only) stat.

PR #201, opened 2021-05-26
Added iframes custom_metric

kevinfarrugia:
This is a custom metric that retrieves the data related to `<iframe>` elements present on the page, including their attributes.

It will be used in the 2021 Resource Hints chapter to retrieve lazy loaded iframes.

**Sample WPT**
https://webpagetest.org/custom_metrics.php?test=210527_AiDcER_7ac78943534ff0d08867441fd07feb36&run=1&cached=0

**Data**
```
{
  "iframes": {
    "iframes": {
      "total": 2,
      "nodes": [
        {
          "tagName": "iframe",
          "loading": "lazy",
          "width": "560",
          "height": "315",
          "src": "https://www.youtube.com/embed/XBfYPp6KF2g",
          "frameborder": "0",
          "allow": "accelerometer; autoplay; encrypted-media; gyroscope; picture-in-picture",
          "allowfullscreen": ""
        },
        {
          "tagName": "iframe",
          "loading": "lazy",
          "width": "560",
          "height": "315",
          "src": "https://www.youtube.com/embed/rrVDATvUitA",
          "frameborder": "0",
          "allow": "accelerometer; autoplay; encrypted-media; gyroscope; picture-in-picture",
          "allowfullscreen": ""
        }
      ],
      "attribute_usage_count": {
        "loading": 2,
        "width": 2,
        "height": 2,
        "src": 2,
        "frameborder": 2,
        "allow": 2,
        "allowfullscreen": 2
      }
    },
    "loading_values": ["lazy", "lazy"]
  }
}
```


custom_metrics/iframes.js
```diff
@@ -0,0 +1,181 @@
+//[iframes]
+// Uncomment the previous line for testing on webpagetest.org
+
+// Sanitize the `attributes` property.
+function getNodeAttributes(node) {
```
rviscomi, line 5: Given the duplication of code in [almanac.js](https://github.com/HTTPArchive/legacy.httparchive.org/blob/master/custom_metrics/almanac.js) would you be open to adding this custom metric there as a new `resource_hints` property of the response object? You could also include #202 in there as well.

I'd prefer not to maintain core functionality like these functions across separate files.

rviscomi (changes requested):

custom_metrics/iframes.js
kevinfarrugia: Updated. Also updated description with new WPT link and JSON.

rviscomi (approved):

PR #202, opened 2021-05-26
Added lazy-in-viewport custom metric

kevinfarrugia:
Updated the Images custom metric to include two additional properties, `loading` - which retrieves the value of the `loading` attribute on the `` element and `in-viewport` - which returns a boolean value indicating whether the image is visible in the viewport.

Will be used in 2021 Resource Hints chapter.

Source: https://docs.webpagetest.org/custom-metrics/examples/lazy-images-in-viewport/

**Sample WPT**
https://webpagetest.org/custom_metrics.php?test=210602_BiDcM8_940d25ae2138a2d2dae32c111fb964cc&run=1&cached=0

```
[
  {
    "url": "https://placekitten.com/401/401",
    "width": 401,
    "height": 401,
    "naturalWidth": 401,
    "naturalHeight": 401,
    "loading": "lazy",
    "in-viewport": true
  },
  {
    "url": "https://placekitten.com/402/402",
    "width": 402,
    "height": 402,
    "naturalWidth": 402,
    "naturalHeight": 402,
    "loading": "lazy",
    "in-viewport": false
  }
]
```


custom_metrics/lazy-in-viewport.js
```diff
@@ -0,0 +1,11 @@
+// [lazy-in-viewport]
+// Uncomment the previous line for testing on webpagetest.org
+
+let images = document.querySelectorAll("img[loading=lazy]");
+let lazyImages = [];
+images.forEach((img) => {
+  if (img.getBoundingClientRect().top < window.innerHeight) {
+    lazyImages.push(img.src);
```
rviscomi, line 8: How do you plan on using the `src` in your analysis? I wonder if it'd be useful to have other metadata about the image.

Also not sure whether this metric should live in its own JS file or combined with another related one, either `almanac.js` or we could start `resource_hints.js` and keep adding to it. WDYT?

kevinfarrugia: I don't need the `src` actually and would be happy with a count (`array.length`). This way I would be able to say

> 10% of all lazy loaded images are actually within the initial viewport.

kevinfarrugia: > Also not sure whether this metric should live in its own JS file or combined with another related one, either almanac.js or we could start resource_hints.js and keep adding to it. WDYT?

I can add it to the `images` custom metric already being used in `almanac.js`. That's your call. I was initially told that separate files are preferred, but either works for me.

kevinfarrugia: I think WPT includes a custom metric by default which includes the `src` of each image and seems to work well with `picture` too.

```
[
  {
    "url": "https://googlechrome.github.io/samples/picture-element/images/kitten-large.png",
    "width": 650,
    "height": 285,
    "naturalWidth": 650,
    "naturalHeight": 285
  },
  {
    "url": "https://googlechrome.github.io/samples/picture-element/images/butterfly.webp",
    "width": 320,
    "height": 320,
    "naturalWidth": 320,
    "naturalHeight": 320
  }
]
```

I think we should be able to use it to answer questions on the file type, domain & whether it has been preloaded. Unless you want to segment this data for lazy-loaded images.

bazzadp: There are no default custom metrics. This is 
 this one: https://github.com/HTTPArchive/legacy.httparchive.org/blob/master/custom_metrics/Images.js

I think you should you add `loading` attributes to that as well as an `in-viewport` boolean, rather than defining a new one?

bazzadp: Actually I see we copied that Images.js from WPT and that still runs on the public instances.

@rviscomi / @pmeenan any thoughts on whether we should leave that alone then so we're still in sync, and define Almanac specific changes to a separate file (either the existing almanac.js or a new resource-hints.js)? Might be confusing if we diverge image.js...

rviscomi: I'd be ok modifying the images metric, parity with WPT isn't a requirement. @pmeenan any other thoughts?

pmeenan: The WPT metric is just for the Cloudinary image analysis integration which doesn't really apply to the HTTP Archive so feel free to modify it or completely replace it.

kevinfarrugia: Based on the above I have modified the Images custom metric and updated the first comment with a link to the updated test and data.

Note that there is a bit of overlap between the Images custom metric and the `images` property in the `almanac.js`. I have not updated the latter.

custom_metrics/Images.js
```diff
@@ -2,19 +2,26 @@ var wptImages = function(win) {
...
-          images.push({'url': url, 'width': el.width, 'height': el.height, 'naturalWidth': el.naturalWidth, 'naturalHeight': el.naturalHeight});
+        if (url.indexOf("http") === 0) {
+          images.push({
+            url: url,
+            width: el.width,
+            height: el.height,
+            naturalWidth: el.naturalWidth,
+            naturalHeight: el.naturalHeight,
+            loading: el.getAttribute("loading"),
+            "in-viewport": el.getBoundingClientRect().top < window.innerHeight,
```
rviscomi, line 20: Would you also like to add a property that combines both values, for example
```suggestion
            "in-viewport": el.getBoundingClientRect().top < window.innerHeight,
            lazyAboveFold: el.getAttribute("loading") == "lazy" && el.getBoundingClientRect().top < window.innerHeight
```

Alternatively, you can evaluate that in the SQL but this would save a step in the analysis.

rviscomi (approved):

custom_metrics/Images.js
kevinfarrugia: Personally I thought it was better to evaluate this in SQL to avoid having multiple metrics (or a combination of them) giving the same information (wasteful) and we may want to change/extend what we are reporting. However I think you are in a better position to know which is the better option.

rviscomi: SGTM

rviscomi, line 20: @kevinfarrugia a suggestion by Colin Bendell in the webperf Slack was to also consider the `left` property to handle carousels that shift images offscreen. Would you be able to open a new PR to update the script, taking that into consideration? Another idea is to use IntersectionObserver but that might be overkill.

colinbendell: I'd also suggest looking at all corners of the image to see if they are within the viewport and not negatively offset out of the top or left

kevinfarrugia: Yes sure, could be done. Will try have a PR by Monday although I think I will immediately opt for the IntersectionObserver as this would probably be a more robust solution.

Issue #203, opened 2021-05-29
Create new scripts to detect importScripts() and usage of SW methods inside pwa.js


demianrenzulli:
Hi folks,

As you know, for the 2021 Web Almanac PWA section, we need some additional metrics, to cover two use cases:

1. Detect all the calls to `importScripts()`, so we can replace the [2020 query](https://github.com/HTTPArchive/almanac.httparchive.org/blob/main/sql/2020/pwa/popular_pwa_libraries.sql) used to create a ranking of popular libraries used by service workers.
2. Getting all the service worker events, to replace the [2020 SW Event Query](https://github.com/HTTPArchive/almanac.httparchive.org/blob/main/sql/2020/pwa/sw_events.sql).

For (1), I added the following code to [pwa.js](https://github.com/HTTPArchive/legacy.httparchive.org/blob/master/custom_metrics/pwa.js), and could obtain the files that contain calls to `importScripts()`. The code is very similar to Rick's code for Workbox detection:

```js
const importScriptsPattern = /importScripts\(.*\);/g;
const importScriptsInfo = response_bodies.filter(har => {
    return importScriptsPattern.test(har.response_body);
}).map(har => {
    return [har.url, Array.from(har.response_body.matchAll(importScriptsPattern)).map(m => m[0])];
});

return {
    serviceWorkers: Object.fromEntries(serviceWorkers),
    manifests: Object.fromEntries(manifests),
    serviceWorkerInitiated: Object.keys(Object.fromEntries(serviceWorkerInitiated)),
    workboxInfo: Object.fromEntries(workboxInfo),
    importScriptsInfo: Object.fromEntries(importScriptsInfo)
};
```
You can take a look at [this test](https://www.webpagetest.org/result/210529_BiDcXD_4b97c175d9af18c19da136cb2f6d510e/?f=json) for https://www.naranja.com. Output at `$.data.runs[1].firstView.pwa.importScriptsInfo`.

Besides that, I'm a bit concerned about the cases where the original script might not detect a service worker (minification might be one cause, as Rick mentioned [here](https://github.com/HTTPArchive/legacy.httparchive.org/pull/198#issue-641825332)).

Here are some example sites that have service workers, where the test returns it empty:

- Twitter (https://mobile.twitter.com): [Test](https://www.webpagetest.org/result/210529_BiDcR8_e0a19861a701779ef6da96553f936ee7/?f=json).
[...]

PR #204, opened 2021-05-30
New script to detect importScripts(), service worker events and service worker properties

demianrenzulli:
Hi folks,

As discussed in #203 I decided to create a PR for this script (even when is not optimal), since, after analyzing some tests with @tunetheweb, having it merged for Monday's crawl could help us get an idea of the number of false negatives in `serviceWorkerInitiatedURLs`.

We have added three new fields to the response: `importScriptsInfo`, `swEventListenersInfo`, `wPropertiesInfo`, which use [Rick's script for Workbox detection](https://github.com/HTTPArchive/legacy.httparchive.org/blob/a6b419f65194abb81840be6e52cfbaedfb945b40/custom_metrics/pwa.js#L72) and combines it with the regular expressions used last year to detect SW events and properties ([sw_events.sql](https://github.com/HTTPArchive/almanac.httparchive.org/blob/a883df882114af3b641791f787c3359bede5d803/sql/2020/pwa/sw_events.sql#L15)), plus a new one to detect calls to `importScripts()`.

### Tests

- Here is a [test](https://www.webpagetest.org/result/210530_AiDcS8_e2a3c376a59afc821bf38a203c510922/?f=json) for https://www.naranja.com. In this case the `serviceWorkers` field is non-empty and it also has values for the new properties.
- Here is another [test](https://www.webpagetest.org/result/210530_AiDcX2_781fbed981df2fb15d8270d9800afebf/?f=json) for https://mobile.twitter.com, where the `serviceWorkers` is empty, but we can see `swEventListenersInfo` and `wPropertiesInfo` populated.

It would be interesting to see how many cases like Twitter we have, to see if we are under-counting many sites that use service workers.

From what we could check with @tunetheweb this number represent a very small percentage.

It would be great if @rviscomi or @jeffposnick can take a look at some point.

cc // @obto

### DIFF
--- a/custom_metrics/pwa.js
+++ b/custom_metrics/pwa.js
@@ -70,11 +70,13 @@ const serviceWorkerInitiatedURLs = new Set(Array.from(serviceWorkerURLs).flatMap
 const serviceWorkerInitiated = getEntriesForURLs(serviceWorkerInitiatedURLs);
 
 // We should use serviceWorkerInitiatedURLs here, but SW detection has some false negatives.
-function getInfoForPattern(regexPattern) {
+function getInfoForPattern(regexPattern, extractMatchingGroupOnly) {
   return response_bodies.filter(har => {
     return regexPattern.test(har.response_body);
   }).map(har => {
-    return [har.url, Array.from(har.response_body.matchAll(regexPattern)).map(m => m[0])];
+    return [har.url, Array.from(har.response_body.matchAll(regexPattern)).map(m => {
+      return extractMatchingGroupOnly ? m[1] : m[0]
+    })];
   });
 }
 
@@ -85,10 +87,10 @@ const importScriptsPattern = /importScripts\(.*\);/g;
 const importScriptsInfo = getInfoForPattern(importScriptsPattern);
 
 const swEventListenersPattern = /addEventListener\(\s*[\'"](install|activate|fetch|push|notificationclick|notificationclose|sync|canmakepayment|paymentrequest|message|messageerror|periodicsync|backgroundfetchsuccess|backgroundfetchfailure|backgroundfetchabort|backgroundfetchclick)[\'"]/g;
-const swEventListenersInfo = getInfoForPattern(swEventListenersPattern);
+const swEventListenersInfo = getInfoForPattern(swEventListenersPattern, true);
 
 const swPropertiesPattern = /\.on(install|activate|fetch|push|notificationclick|notificationclose|sync|canmakepayment|paymentrequest|message|messageerror|periodicsync|backgroundfetchsuccess|backgroundfetchfailure|backgroundfetchabort|backgroundfetchclick)\s*=/g;
-const swPropertiesInfo = getInfoForPattern(swPropertiesPattern);
+const swPropertiesInfo = getInfoForPattern(swPropertiesPattern, true);
 
 const swMethodsPattern = /skipWaiting\(\)|navigationPreload.(enable|disable|setHeaderValue|getState)/g;
 const swMethodsInfo = getInfoForPattern(swMethodsPattern);
--- a/custom_metrics/pwa.js
+++ b/custom_metrics/pwa.js
@@ -107,6 +107,10 @@ const windowEventListenersInfo = getInfoForPattern(windowEventListenersPattern, 
 const windowPropertiesPattern = /\.on(appinstalled|beforeinstallprompt)\s*=/g;
 const windowPropertiesInfo = getInfoForPattern(windowPropertiesPattern, true);
 
+function isObjectKeyEmpty(field) {
+  return field == null || field.length == 0;
+}
+
 return {
   serviceWorkers: Object.fromEntries(serviceWorkers),
   manifests: Object.fromEntries(manifests),
@@ -119,5 +123,7 @@ return {
   swObjectsInfo: Object.fromEntries(swObjectsInfo),
   swRegistrationPropertiesInfo: Object.fromEntries(swRegistrationPropertiesInfo),
   windowEventListenersInfo: Object.fromEntries(windowEventListenersInfo),
-  windowPropertiesInfo: Object.fromEntries(windowPropertiesInfo)
+  windowPropertiesInfo: Object.fromEntries(windowPropertiesInfo),
+  //Experimental field: Heuristic to detect if a site has a service worker even if the 'serviceWorkers' field is empty (false positives).
+  serviceWorkerHeuristic: !isObjectKeyEmpty(serviceWorkers) || !isObjectKeyEmpty(workboxInfo) || !isObjectKeyEmpty(importScriptsInfo) || !isObjectKeyEmpty(swEventListenersInfo) || !isObjectKeyEmpty(swMethodsInfo)
 };
--- a/custom_metrics/pwa.js
+++ b/custom_metrics/pwa.js
@@ -90,6 +90,12 @@ const swEventListenersInfo = getInfoForPattern(swEventListenersPattern);
 const swPropertiesPattern = /\.on(install|activate|fetch|push|notificationclick|notificationclose|sync|canmakepayment|paymentrequest|message|messageerror|periodicsync|backgroundfetchsuccess|backgroundfetchfailure|backgroundfetchabort|backgroundfetchclick)\s*=/g;
 const swPropertiesInfo = getInfoForPattern(swPropertiesPattern);
 
+const swMethodsPattern = /skipWaiting\(\)|navigationPreload.(enable|disable|setHeaderValue|getState)/g;
+const swMethodsInfo = getInfoForPattern(swMethodsPattern);
+
+const swObjectsPattern = /clients\.(get|matchAll|openWindow|claim)|client\.(postMessage|id|type|url)|caches\.(match|has|open|delete|keys)|cache\.(match|matchAll|add|addAll|put|keys)/g;
+const swObjectsInfo = getInfoForPattern(swObjectsPattern);
+
 return {
   serviceWorkers: Object.fromEntries(serviceWorkers),
   manifests: Object.fromEntries(manifests),
@@ -97,5 +103,7 @@ return {
   workboxInfo: Object.fromEntries(workboxInfo),
   importScriptsInfo: Object.fromEntries(importScriptsInfo),
   swEventListenersInfo: Object.fromEntries(swEventListenersInfo),
-  swPropertiesInfo: Object.fromEntries(swPropertiesInfo)
+  swPropertiesInfo: Object.fromEntries(swPropertiesInfo),
+  swMethodsInfo: Object.fromEntries(swMethodsInfo),
+  swObjectsInfo: Object.fromEntries(swObjectsInfo)
 };
--- a/custom_metrics/pwa.js
+++ b/custom_metrics/pwa.js
@@ -75,7 +75,7 @@ function getInfoForPattern(regexPattern, extractMatchingGroupOnly) {
     return regexPattern.test(har.response_body);
   }).map(har => {
     return [har.url, Array.from(har.response_body.matchAll(regexPattern)).map(m => {
-      return extractMatchingGroupOnly ? m[1] : m[0]
+      return (extractMatchingGroupOnly && m.length > 0) ? m[1] : m[0]
     })];
   });
 }
@@ -83,8 +83,8 @@ function getInfoForPattern(regexPattern, extractMatchingGroupOnly) {
 const workboxPattern = /(?:workbox:[a-z\-]+:[\d.]+|workbox\.[a-zA-Z]+\.?[a-zA-Z]*)/g;
 const workboxInfo = getInfoForPattern(workboxPattern);
 
-const importScriptsPattern = /importScripts\(.*\);/g;
-const importScriptsInfo = getInfoForPattern(importScriptsPattern);
+const importScriptsPattern = /importScripts\(([^)]*)\)/g;
+const importScriptsInfo = getInfoForPattern(importScriptsPattern, true);
 
 const swEventListenersPattern = /addEventListener\(\s*[\'"](install|activate|fetch|push|notificationclick|notificationclose|sync|canmakepayment|paymentrequest|message|messageerror|periodicsync|backgroundfetchsuccess|backgroundfetchfailure|backgroundfetchabort|backgroundfetchclick)[\'"]/g;
 const swEventListenersInfo = getInfoForPattern(swEventListenersPattern, true);
--- a/custom_metrics/pwa.js
+++ b/custom_metrics/pwa.js
@@ -77,9 +77,33 @@ const workboxInfo = response_bodies.filter(har => {
   return [har.url, Array.from(har.response_body.matchAll(workboxPattern)).map(m => m[0])];
 });
 
+const importScriptsPattern = /importScripts\(.*\);/g;
+const importScriptsInfo = response_bodies.filter(har => {
+  return importScriptsPattern.test(har.response_body);
+}).map(har => {
+  return [har.url, Array.from(har.response_body.matchAll(importScriptsPattern)).map(m => m[0])];
+});
+
+const swEventListenersPattern = /addEventListener\(\s*[\'"](install|activate|fetch|push|notificationclick|notificationclose|sync|canmakepayment|paymentrequest|message|messageerror|periodicsync|backgroundfetchsuccess|backgroundfetchfailure|backgroundfetchabort|backgroundfetchclick)[\'"]/g;
+const swEventListenersInfo = response_bodies.filter(har => {
+  return swEventListenersPattern.test(har.response_body);
+}).map(har => {
+  return [har.url, Array.from(har.response_body.matchAll(swEventListenersPattern)).map(m => m[0])];
+});
+
+const swPropertiesPattern = /\.on(install|activate|fetch|push|notificationclick|notificationclose|sync|canmakepayment|paymentrequest|message|messageerror|periodicsync|backgroundfetchsuccess|backgroundfetchfailure|backgroundfetchabort|backgroundfetchclick)\s*=/g;
+const wPropertiesInfo = response_bodies.filter(har => {
+  return swEventListenersPattern.test(har.response_body);
+}).map(har => {
+  return [har.url, Array.from(har.response_body.matchAll(swPropertiesPattern)).map(m => m[0])];
+});
+
 return {
   serviceWorkers: Object.fromEntries(serviceWorkers),
   manifests: Object.fromEntries(manifests),
   serviceWorkerInitiated: Object.keys(Object.fromEntries(serviceWorkerInitiated)),
-  workboxInfo: Object.fromEntries(workboxInfo)
+  workboxInfo: Object.fromEntries(workboxInfo),
+  importScriptsInfo: Object.fromEntries(importScriptsInfo),
+  swEventListenersInfo: Object.fromEntries(swEventListenersInfo),
+  wPropertiesInfo: Object.fromEntries(wPropertiesInfo)
 };
--- a/custom_metrics/pwa.js
+++ b/custom_metrics/pwa.js
@@ -69,34 +69,26 @@ const manifests = getEntriesForURLs(manifestURLs).map(([url, body]) => {
 const serviceWorkerInitiatedURLs = new Set(Array.from(serviceWorkerURLs).flatMap(getURLsInitiatedBy));
 const serviceWorkerInitiated = getEntriesForURLs(serviceWorkerInitiatedURLs);
 
-const workboxPattern = /(?:workbox:[a-z\-]+:[\d.]+|workbox\.[a-zA-Z]+\.?[a-zA-Z]*)/g;
 // We should use serviceWorkerInitiatedURLs here, but SW detection has some false negatives.
-const workboxInfo = response_bodies.filter(har => {
-  return workboxPattern.test(har.response_body);
-}).map(har => {
-  return [har.url, Array.from(har.response_body.matchAll(workboxPattern)).map(m => m[0])];
-});
+function getInfoForPattern(regexPattern) {
+  return response_bodies.filter(har => {
+    return regexPattern.test(har.response_body);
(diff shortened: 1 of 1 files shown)


custom_metrics/pwa.js
```diff
@@ -69,17 +69,33 @@ const manifests = getEntriesForURLs(manifestURLs).map(([url, body]) => {
...
+const workboxInfo = getInfoForPattern(workboxPattern);
+
+const importScriptsPattern = /importScripts\(.*\);/g;
+const importScriptsInfo = getInfoForPattern(importScriptsPattern);
+
+const swEventListenersPattern = /addEventListener\(\s*[\'"](install|activate|fetch|push|notificationclick|notificationclose|sync|canmakepayment|paymentrequest|message|messageerror|periodicsync|backgroundfetchsuccess|backgroundfetchfailure|backgroundfetchabort|backgroundfetchclick)[\'"]/g;
+const swEventListenersInfo = getInfoForPattern(swEventListenersPattern);
+
+const swPropertiesPattern = /\.on(install|activate|fetch|push|notificationclick|notificationclose|sync|canmakepayment|paymentrequest|message|messageerror|periodicsync|backgroundfetchsuccess|backgroundfetchfailure|backgroundfetchabort|backgroundfetchclick)\s*=/g;
+const swPropertiesInfo = getInfoForPattern(swPropertiesPattern);
```
rviscomi, line 91: Do you need to distinguish between `addEventListener` and `on` usage in your analysis? If not, consider combining the regexes or merging the results into a single output property.

Also FYI we have the new [event-names.js](https://github.com/HTTPArchive/legacy.httparchive.org/blob/master/custom_metrics/event-names.js) custom metric that will extract any event name after `addEventListener`. Probably still good to have a shortlist of SW-related event names to check for here but just wanted to make you aware in case it's useful down the road.

rviscomi (changes requested):

demianrenzulli:
Hi @rviscomi! Thank you for taking a look at this PR.

I'll be running some tests, to decide what's the best course of action for both suggestions, according to what the data returns and how we visualize it.

One important question:

I've noticed that the original script takes into account all the URLs (`response_bodies`), not only the service worker ones (`serviceWorkerInitiated`).

For Workbox-related things or `importScripts()` this is fine, since these strings won't appear outside service workers. In other cases, like the `onmessage` property or `message` event, there can be a lot of instances of them outside service worker files. This year I was planning to also include other service worker related objects, like [Cache](https://developer.mozilla.org/en-US/docs/Web/API/Cache), that can also be accessed client-side, leading to stats that might not necessarily be related to service workers usage or PWAs.

Correct me if I'm wrong, but I believe one way of mitigating this, would be to use `serviceWorkerInitiated` instead of `response_bodies`, but as you've mentioned [here](https://github.com/HTTPArchive/legacy.httparchive.org/blob/a6b419f65194abb81840be6e52cfbaedfb945b40/custom_metrics/pwa.js#L73), there might be many false positives, like the case of [the Twitter test](https://www.webpagetest.org/result/210530_AiDcX2_781fbed981df2fb15d8270d9800afebf/?f=json) I shared before.

Now, if based on the data we receive from the June crawl, we are able to determine that these false positives are relatively low (for example, by comparing how many tests have an empty `serviceWorkers` field, and a non-empty: `workboxInfo`), or once this PR is merged, by checking tests with an empty `serviceWorkers` field and a non-empty `importScriptsInfo` / `swEventListenersInfo` / `swPropertiesInfo`, would it be more advisable to run the scripts on `serviceWorkerInitiated` instead?

I believe 2019 and 2020 were taking the whole `response_bodies` into account, so I don't want to introduce a change that might produce an intended impact on the data. Also, since we couldn't introduce these new metrics, we might not have a lot of time to make that decision.

Thoughts are welcome!

cc // @bazzadp

demianrenzulli:
Also, this comment from @bazzadp is very interesting: 

> WPT also detected they were service worker calls as they are in blue. So could also ask @pmeenan how it does that and use that potentially?

For the [Twitter test](https://www.webpagetest.org/video/compare.php?tests=210529_AiDc3P_da59493935913429572a97b5372948d0-r:1-c:0), that we couldn't detect the service worker with our script, WPT shows the service worker related requests in blue in the waterfall.

image

Maybe we can reuse that logic WPT uses to detect service worker activity inside our script to be sure of which tests things from a SW?

Cheers.

pmeenan:
WPT highlights any request that belongs to a different document from the main page in blue.  It will show SW but also iFrames.  "documentURL" in the request data.

demianrenzulli:
Thanks for confirming @pmeenan! In that case, we won't be able to use that to isolate SW-only traffic.

rviscomi:
> Correct me if I'm wrong, but I believe one way of mitigating this, would be to use serviceWorkerInitiated instead of response_bodies, but as you've mentioned here, there might be many false positives, like the case of the Twitter test I shared before.

That's right. It would provide fewer false positives but we'd lose out on the Twitter edge case.

> Now, if based on the data we receive from the June crawl, we are able to determine that these false positives are relatively low (for example, by comparing how many tests have an empty serviceWorkers field, and a non-empty: workboxInfo), or once this PR is merged, by checking tests with an empty serviceWorkers field and a non-empty importScriptsInfo / swEventListenersInfo / swPropertiesInfo, would it be more advisable to run the scripts on serviceWorkerInitiated instead?

Reevaluating after the June crawl sounds like a good plan. We can also compare to the ServiceWorkerControlledPage use counter.

tunetheweb:
> We can also compare to the ServiceWorkerControlledPage use counter.

@demianrenzulli and I looked at this for 2020 stats and it was pretty damn close actually.

Web Almanac counted 53,366 sites for mobile and 42,521 sites for desktop.

Blink Usage for Aug 2020 had 55,019 and 49,305. And they presumably will be counting more than the home page and all sites not just CrUX ones?

So that’s VERY close in the grand scheme of things and gives real confidence the 2020 methodology (which is what we are currently proposing for 2021 but just moving the counting to be done during the crawl) is representative.

tunetheweb:
Actually where does the `httparchive.blink_usage` data come from? Is that from our HTTPArchive crawl (so will be just home pages after all) or from Chrome’s stats?

pmeenan:
The feature usage comes from the crawl.

tunetheweb:
> The feature usage comes from the crawl.

Ah then the closeness is more understandable 😔

I guess the difference would be sites like Twitter where we can’t see it registering a service worker even though it does use one.

rviscomi:
One other consideration for minimizing false negatives (eg onmessage outside of a SW) is to test for patterns that would disqualify the script from being a service worker. For example, if we find `document.querySelector` in the script with the `onmessage` event listener, then it's likely that the script isn't a SW. WDYT would that kind of approach work, and if so what kinds of signals can you think of that would strongly indicate that we should discount the script?

demianrenzulli:
Thanks for sharing your thoughts again, @rviscomi! It's really helping us understand what can be done to make our analysis more precise.

So far, I have found a potentially strong usage of features outside service worker in the following cases:

- The [MessageEvent](https://developer.mozilla.org/en-US/docs/Web/API/ServiceWorkerGlobalScope/message_event).
- The [Caches Interface](https://developer.mozilla.org/en-US/docs/Web/API/ServiceWorkerGlobalScope/caches).
- The [fetch() method](https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/fetch).

All of them can be accessed both from pages and service workers, and their usage might be quite frequent in both contexts. So, we either (1) exclude them from our analysis to avoid false positives, or (2) find a way of mitigating this issue by focusing only in `serviceWorkerInitiated` or something similar.

For (1): For the three features, I thought it might be interesting to know how many service workers use the `message` event listener, since that could let us know how many sites are using "Page to Service worker" communication techniques. Excluding `fetch()` or `caches`, might not be so bad, in my opinion. I assume that most service workers use both of these features, so I don't think that knowing that will give us very interesting insights.

Regarding (2): If we decide to implement a technique to filter service worker only activity, this comment:

> For example, if we find document.querySelector in the script with the onmessage event listener, then it's likely that the script isn't a SW. WDYT would that kind of approach work, and if so what kinds of signals can you think of that would strongly indicate that we should discount the script?

Sounds like a very good idea. In general, workers don't have access to the [Window](https://developer.mozilla.org/en-US/docs/Web/API/Window) and therefore the [Document](https://developer.mozilla.org/en-US/docs/Web/API/Document) object either and have limited access to browser APIs.

Since Window is implicit, I believe that any call to `document.*` (like the example you've shared `document.querySelector`),   should be forbidden inside worker contexts. 

I'm wondering how this could be implemented? Is it possible to create an exception in pwa.js, or is it something we can do after getting the data?

demianrenzulli:
BTW: cc @jeffposnick to validate this:

> Sounds like a very good idea. In general, workers don't have access to the Window and therefore the Document object either and have limited access to browser APIs.
> 
> Since Window is implicit, I believe that any call to document.* (like the example you've shared document.querySelector), should be forbidden inside worker contexts.

jeffposnick:
That's correct. `document` isn't part of the `ServiceWorkerGlobalScope`.

custom_metrics/pwa.js
```diff
@@ -69,17 +69,33 @@ const manifests = getEntriesForURLs(manifestURLs).map(([url, body]) => {
...
-const workboxPattern = /(?:workbox:[a-z\-]+:[\d.]+|workbox\.[a-zA-Z]+\.?[a-zA-Z]*)/g;
 // We should use serviceWorkerInitiatedURLs here, but SW detection has some false negatives.
-const workboxInfo = response_bodies.filter(har => {
-  return workboxPattern.test(har.response_body);
-}).map(har => {
-  return [har.url, Array.from(har.response_body.matchAll(workboxPattern)).map(m => m[0])];
-});
+function getInfoForPattern(regexPattern) {
+  return response_bodies.filter(har => {
+    return regexPattern.test(har.response_body);
```
rviscomi, line 75: Per the discussion, I'd suggest adding a condition here to filter out responses that contain non-SW code. For example, have a top-level const to define a pattern that disqualifies the response from being considered a SW and test for that here:

```suggestion
    return !disqualifyingSWPattern.test(har.response_body) && regexPattern.test(har.response_body);
```

Where `disqualifyingSWPattern` could be defined as something like:

```js
const disqualifyingSWPattern = /\bdocument\.\w+/g;
```

(not tested)

demianrenzulli: This is great! I'm working on a new iteration to add more SW features to detect. I'll update the PR and then work on this condition to see if we can achieve the desired effect.

demianrenzulli:
Hi folks,

I just updated this PR, by adding two additional fields: `swMethodsInfo` and `swObjectsInfo`.
Tests:

- [Twitter](https://www.webpagetest.org/result/210608_BiDc4B_4402d8054f625669f13348bed80e8426/?f=json).
- [Terra](https://www.webpagetest.org/result/210608_BiDcB4_2704fb0fa57de9e8b56c28a2fb544378/?f=json).

Before starting to work on @rviscomi code to filter SW only files, I wanted to go back to on of his previous comments:

> Do you need to distinguish between addEventListener and on usage in your analysis? If not, consider combining the regexes or merging the results into a single output property.

I'm not familiar on how this information will be consumed and if it's better to combine some of the fields that this script returns into one.

I'd like to hear @tunetheweb thoughts on this, before changing the current structure. Maybe we can discuss in tomorrow's sync.

demianrenzulli:
Hi team,

As discussed in our sync with @tunetheweb and @obto earlier this week, I've been working in some updates to the script to get it closer to the final version, so we can use it for July's crawl. @rviscomi (and others) I would appreciate your thoughts:

### New fields

You'll notice that I have include more PWA features to test, like `navigationPreload` and events like `appinstalled`.

### Output format

I added an option to `getInfoForPattern()` to return only the matching group for some patterns. For example: `importScripts(...)` only returns the contained URL. Same for `addEventListener()`, etc.

### Granularity level

As discussed, we would prefer to return the fields in the more granular way, instead of combining them, so I tried to group them according to the feature they belong to.

With that say: `swEventListenersInfo` and `swPropertiesInfo` continue being separate fields, even when they might test similar things.

### Filtering "service worker only" activity

I took @rviscomi's suggestion and tried the suggested technique:

```javascript
const disqualifyingSWPattern = /\bdocument\.\w+/g;
```

While I was initially happy with the preliminary results, I found some cases like [Spotify's service worker](https://open.spotify.com/service-worker.js), which code contains calls to `document`. As discussed this is illegal in this context, but might be the result of using some build tools or polyfills to generate the service worker. 
We discussed this with @jeffposnick offline, and thought that it might be risky to use this pattern as a way to detect SW-only activity.

For that reason, I decided to exclude two features that are not SW-specific and that are very commonly used in pages: `fetch()` and `message`. After that the results are much more clean. 

Something else that we can do after analyzing June's crawl would be creating a condition like: 

_"If it doesn't have a `serviceWorkers` field, but it has SW-specific methods, we can consider that it uses a service worker"._

### Test cases

I tested [20 sites that have service workers](https://docs.google.com/spreadsheets/d/1ibWjh9_3wL6SJ1QW_vVxZiwB2bQQ4DvMhXWVHSAgNSg/edit?resourcekey=0-aLUp0HYyuZVl0Ye9xMLYSw#gid=0) and the output seems to be working well. I was hoping we can use that as test cases every time we introduce changes.

@tunetheweb: you can take a look at any test, for example, [this one](https://www.webpagetest.org/result/210610_AiDcSS_378a25e000712f616bd14f59f47d1142/?f=json), to see how the JSON looks like.

Sorry for the length of this comment! Too many decisions have been made that I wanted to share.

tunetheweb:
Hey @demianrenzulli is [this expected in your results](https://www.webpagetest.org/jsonResult.php?test=210610_AiDcSS_378a25e000712f616bd14f59f47d1142&pretty=1&rv=0&requests=0&standard=0&median=0&runs=1&console=0):

image

Given that you are searching for very specific events how can the `e` get in there?

tunetheweb:
Oh and btw I just found out yesterday that if you use URLs like this: `https://www.webpagetest.org/jsonResult.php?test=210610_AiDcSS_378a25e000712f616bd14f59f47d1142&pretty=1&rv=0&requests=0&standard=0&median=0&runs=1&console=0` with the query params to strips out a lot of the JSON before sending back, it is a lot quicker to load when testing your scripts and doesn't kill your Chrome instance! 😀

demianrenzulli:
Hi @tunetheweb, thank you for taking a look at the tests and share your feedback!

The output you found belongs to an `importScripts()` call inside one of Workbox files:

https://storage.googleapis.com/workbox-cdn/releases/4.3.1/workbox-sw.js

This is the relevant portion of the code:

```javascript
loadModule(t) {
   const e  = this.i(t);
      try {
         importScripts(e), this.o = !0
      } catch (s) {
         throw console.error(`Unable to import module '${t}' from '${e}'.`), s
      }
}
```

As you can see, the regular expression has detected the pattern well, but since the call to `importScripts()` receives a variable as a parameter, it's returning that value as its content.

I didn't want to force the the parameter sent to `importScripts()`to start with `'` or `"`, to avoid excluding potentially valid cases, and also in case we want to come up with an estimation like: "X% of sites use importScripts()".

Maybe at the time of creating the raking of most used libraries we could filter out these results that don't add much?
It seems like this is [this is the query](https://github.com/HTTPArchive/almanac.httparchive.org/blob/main/sql/2020/pwa/popular_pwa_libraries.sql) used last year, so any recommendations you might have are highly welcome.

tunetheweb:
Ah I loaded the wrong script so couldn't see it! My bad.

Is it useful to know *what* they import? Or just that they use `importScript`? Do we need the parameter at all or just the list of scripts that call it?

i.e. instead of this:

```json
"importScriptsInfo": {
    "https: //www.tiktok.com/sw.js": [
        "'https://storage.googleapis.com/workbox-cdn/releases/4.3.1/workbox-sw.js'"
    ],
    "https: //storage.googleapis.com/workbox-cdn/releases/4.3.1/workbox-sw.js": [
        "e"
    ]
}
```

Should we have this?:

```json
"importScriptsInfo": [
    "https: //www.tiktok.com/sw.js",
    "https: //storage.googleapis.com/workbox-cdn/releases/4.3.1/workbox-sw.js"
]
```

Or maybe each just this?:

```json
"importScriptsInfo": true
```

demianrenzulli:
Thanks again @tunetheweb!

If we do this:

```javascript
"importScriptsInfo": [
    "https: //www.tiktok.com/sw.js",
    "https: //storage.googleapis.com/workbox-cdn/releases/4.3.1/workbox-sw.js"
]
```

We know that there are two files that contain `importScripts()`, but we don't know which are the libraries they are importing, so we won't be able to create the ranking of most used libraries. Am I right?

Same goes for `"importScriptsInfo": true`. It would help us know how many sites have `importScripts()`, but not which libraries they are importing.

Please, let me know if I understood this correctly.

tunetheweb:
Ah you're correct yes. We want to track the most used libraries, not just the use of import scripts.

Then maybe check it includes at least one `.`? Obviously this will exclude all the variable type calls to importScripts but they're of limited use in it's current form anyway.

Or we just leave as is and filter out in the results?

WDYT?

demianrenzulli:
I really like this idea:

`Then maybe check it includes at least one .?`, but what if a site has a service worker with only an `importScripts()` that's being loaded by sending it a variable?

While the variable wouldn't be useful for the rankings, not knowing that site has an `importScripts()` call won't let us come up with a realistic number for the % of sites that uses `importScripts()`.

That's the first row in [last year's table](https://almanac.httparchive.org/en/2020/pwa#popular-import-scripts).

Am I right?

tunetheweb:
I say leave as is then and can clean it up in Sheets.

I also had a look at your test cases and looks pretty thorough! My only concern is the number of "Empty serviceWorkers field with non empty SW methods" cases you have. Is that because you specifically looked for them? Or is our logic so limited that it's picking up so few of them? Anything we can do to improve it that you've seen from these sites? Though guess we do have Blink Usage and Lighthouse to fall back on for those tests.

demianrenzulli:
I have to admit that I am a bit concerned too about the potentially large number of false negatives that we have seen in the tests for the `serviceWorkers` field. One hypothesis isthat, since I tested mostly well known sites that might be more sophisticated and use advanced minification techniques, this number is relatively large in this group, but maybe won't be when analyzing all the URLs from the dataset.

If we find that the number of false positives is still considerable, something that we can do is to create a condition like:

"If a site has an empty `serviceWorkers` field but either `importScriptsInfo` or `swEventListenersInfo` or `swMethodsInfo` are not empty, we can assume that it has a service worker".

A priori, none of these fields should be populated for sites that don't have it, since the methods they check for shouldn't appear outside of service workers. But maybe it's good to have the data to be 100% sure.

Please, let me know what you think!

demianrenzulli:
Here is [a test](https://www.webpagetest.org/result/210611_BiDcZQ_8e278056a8dfdb5e9a418f3c07115336/?f=json) for web.dev, which doesn't have a service worker, and therefore, has all the SW-related fields mentioned before, empty.

demianrenzulli:
Hi folks, a final update on my side:

I just committed an experimental field `serviceWorkerHeuristic` that will be `true` if either `serviceWorkers`, `workboxInfo`, `importScriptsInfo`, `swEventListenersInfo` or `swMethodsInfo` are non-empty.

If you take a look at the **SW Heuristics** column of the [test case sheet](https://docs.google.com/spreadsheets/d/1ibWjh9_3wL6SJ1QW_vVxZiwB2bQQ4DvMhXWVHSAgNSg/edit?resourcekey=0-aLUp0HYyuZVl0Ye9xMLYSw#gid=0), you'll see that this value is `true` for all the PWA sites tested there.

I added a new tab in the sheet, **Non-PWA sites**, with URLs for sites that don't register a service worker, and you'll see that in all cases **SW Heuristics** is `false`.

This might help us get a more accurate idea of how many sites actually have service workers, but since the number of tests I performed is relatively small (30 approximately), I can't be sure what the result will be in the order of thousands when the actual crawl takes place.

I hope this helps.

custom_metrics/pwa.js
```diff
@@ -69,17 +69,33 @@ const manifests = getEntriesForURLs(manifestURLs).map(([url, body]) => {
...
+  }).map(har => {
+    return [har.url, Array.from(har.response_body.matchAll(regexPattern)).map(m => m[0])];
```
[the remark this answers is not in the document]
demianrenzulli: I made this change by adding an additional parameter (`extractMatchingGroupOnly`) to the `getInfoForPattern()` function. As you mentioned, in some cases, we want `m[0]` and in others `m[1]`.

custom_metrics/pwa.js
demianrenzulli: We decided not to combine fields, since they are easier to merge during queries, but harder to separate, if, for some reason we need to use the original separation. In this case, for example, to answer the question: "How many sites use event listeners vs. properties".
Please, refer to the **Granularity level** headline in [this comment](https://github.com/HTTPArchive/legacy.httparchive.org/pull/204#issuecomment-858998233) for more details.

custom_metrics/pwa.js
demianrenzulli: Please, refer to the **Filtering "service worker only" activity** of [this comment](https://github.com/HTTPArchive/legacy.httparchive.org/pull/204#issuecomment-858998233), where I explained why this strategy didn't work.
Instead of that, I ended up adding an experimental field `serviceWorkerHeuristic`, which combines different conditions for features that should only appear in service workers. This is explained in [this comment](https://github.com/HTTPArchive/legacy.httparchive.org/pull/204#issuecomment-860066559).

demianrenzulli:
@rviscomi as discussed with @tunetheweb  I’ve resolved all the comments and don’t plan on making any further changes here if you can review again when you have a moment?

I did my best to cover [as many test cases as possible](https://docs.google.com/spreadsheets/d/1ibWjh9_3wL6SJ1QW_vVxZiwB2bQQ4DvMhXWVHSAgNSg/edit?resourcekey=0-aLUp0HYyuZVl0Ye9xMLYSw#gid=0).

rviscomi (approved):
Looks great!

PR #205, opened 2021-06-07
Creation of metric summary file

obto:
To help analysts get a quick overview of all the metrics we are tracking we will start providing summaries of each metric contained in each custom metrics file

### DIFF
--- a/custom_metrics/metric-summary.md
+++ b/custom_metrics/metric-summary.md
@@ -115,7 +115,7 @@ Example response:
 
 ## 10.5
 
-*DEPRECATED*
+*DEPRECATED*. [See wpt_bodies.js](https://github.com/HTTPArchive/legacy.httparchive.org/blob/master/custom_metrics/wpt_bodies.js)
 
 Extracts schema.org elements and finds all @context and @type usage.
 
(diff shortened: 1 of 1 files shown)


custom_metrics/metric-summary.md
```diff
@@ -0,0 +1,706 @@
...
+    "name": 8,
+    "content": 15,
+    "property": 7
+  }
+}
+```
+
+## 10.5
+
+*DEPRECATED*
```
rviscomi, line 118: How should we think about deprecated metrics? Are there always better replacements somewhere else (if so where) or should we delete them?

rviscomi (commented):
This is awesome, thanks for documenting all of this! Just one question about deprecated metrics.

custom_metrics/metric-summary.md
obto: Yes, both of those deprecated metrics are in the SEO chapters metrics. Once the wpt_bodies.js file is written in here, we can link that the new version can be found there.

rviscomi: Ok SGTM. Could you add a link to that JS file so if someone is looking to query this data and finds this documentation first, they know where to look instead? Once we have docs for the other file it'll be easier to link together but we can link to the source JS itself in the mean time.

obto: Done

rviscomi (approved):

PR #209, opened 2021-06-18
[Custom Fugu metrics] Add Gamepad detection

tomayac:
(Sorry, forgot about this one.)

### DIFF
--- a/custom_metrics/fugu-apis.js
+++ b/custom_metrics/fugu-apis.js
@@ -213,8 +213,11 @@ const patterns = {
   },
 };
 
-// To make sure we don't match on, e.g., blog posts that contain the patterns,
-// make sure that the file names fulfill certain conditions as a heuristic.
+// To avoid to match on, e.g., blog posts that contain the patterns,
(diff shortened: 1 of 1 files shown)


tunetheweb:
BTW meant to comment on the last one. Any concern this won't pick up inline `<SCRIPT>` blocks in HTML files?

I imagine usage of those for these more advanced APIs is **VERY small** and the risk of picking up false positives greatly increases when including HTML for the documentation usage case you give, so right call I thin. But maybe worth making an explicit comment about this? Especially if anyone else bases their custom metrics on your code where they might not want the same decision if inline usage is more likely.

i.e. add last two lines below:

```
// To make sure we don't match on, e.g., blog posts that contain the patterns,
// make sure that the file names fulfill certain conditions as a heuristic.
// Note this leaves slight risk of excluding inline <script> elements using these APIs
// but usage there is expected to be small and prefer to avoid risk of false positives.
```

tomayac:
This is indeed on purpose, I'll add your proposed comments.

tunetheweb (approved):

PR #210, opened 2021-06-18
Some refinements to the Fugu custom metrics

tomayac:
- Differentiate WebOTP and Credential Management. 
- Add more data to the object.


rviscomi:
Trying to imagine what the new properties will be used for. Is `supported` a way to see when these capabilities land in Chrome stable (the browser used for HTTP Archive testing in WebPageTest)?

tomayac:
These new fields might come in handy for the _future_ of this report to feature-detect support. The field `supported` contains a function that returns a promise that resolves with `true` or `false` dependent on the support situation. The field `featureDetection` contains the function's source code, so it could be used in an `eval()`.

PR #211, opened 2021-06-20
Privacy 2021 custom metrics

max-ostapenko:
Progress on https://github.com/HTTPArchive/almanac.httparchive.org/issues/2149

- [x] IAB TCF v1
- [x] IAB TCF v2
- [x] US Privacy API
- [x] FLOC, interestCohort
- [x] DoNotTrack
- [x] GlobalPrivacyControl
- [x] Ads Transparency Spotlight (top frame check)
- [x] Privacy policy keywords
- [x] Sensitive resources
  - [x] permissions policy
  - [x] media devices
  - [x] geolocation
- [x] Origin trials
  - https://webpagetest.org/result/210629_AiDc2Y_5832ddb0a227fcc48ba1963737b6f324/
- [x] Referrer-Policy
- [x] Feature policy


custom_metrics/privacy.js
```diff
@@ -1,24 +1,141 @@
...
-    var pattern = new RegExp(words, 'i');
+    var pattern = new RegExp(words, 'i');https://floc.glitch.me/
```
max-ostapenko, line 36: @VictorLeP should this URL be here?
```suggestion
    var pattern = new RegExp(words, 'i');
```

custom_metrics/privacy.js
```diff
@@ -1,24 +1,141 @@
...
+  navigator_doNotTrack: findStringInBodies('navigator.doNotTrack'),
+  doNotTrack: findStringInBodies('doNotTrack'),
+  // GPC - test site: https://global-privacy-control.glitch.me/
+  navigator_globalPrivacyControl: findStringInBodies('navigator.globalPrivacyControl'),
+  globalPrivacyControl: findStringInBodies('globalPrivacyControl'),
+  // sensitive resources - accesses permissions policy
+  document_permissionsPolicy: findStringInBodies('document.permissionsPolicy'),
+  permissionsPolicy: findStringInBodies('permissionsPolicy'),
+  // user media (camera, microphone)
+  navigator_mediaDevices_getUserMedia: findStringInBodies('navigator.mediaDevices.getUserMedia'),
```
max-ostapenko, line 131: @VictorLeP 
```suggestion
  navigator_mediaDevices_getUserMedia: findStringInBodies('navigator.mediaDevices.enumerateDevices'),
```
'getUserMedia' prompts the user to give permissions to access the stream.
I would rather use `enumerateDevices` which silently reads device info.

custom_metrics/privacy.js
```diff
@@ -1,24 +1,141 @@
...
+      });
+    } else {
+      uspData = null;
+    }
+    return uspData;
+  })(),
+
+  // Ads Transparency Spotlight Data Disclosure schema
+  // Check `meta` tag cf. https://github.com/Ads-Transparency-Spotlight/documentation/blob/main/implement.md
+  // Only for top frame; can't access child frames (same-origin policy)
```
max-ostapenko, line 94: @VictorLeP In what case do we need it? The extension itself also looks in the website's DOM.

custom_metrics/privacy.js
```diff
@@ -1,24 +1,141 @@
...
+  navigator_globalPrivacyControl: findStringInBodies('navigator.globalPrivacyControl'),
+  globalPrivacyControl: findStringInBodies('globalPrivacyControl'),
+  // sensitive resources - accesses permissions policy
+  document_permissionsPolicy: findStringInBodies('document.permissionsPolicy'),
+  permissionsPolicy: findStringInBodies('permissionsPolicy'),
+  // user media (camera, microphone)
+  navigator_mediaDevices_getUserMedia: findStringInBodies('navigator.mediaDevices.getUserMedia'),
+  mediaDevices: findStringInBodies('mediaDevices'),
+  getUserMedia: findStringInBodies('getUserMedia'),
+  // geolocation
```
max-ostapenko, line 134: @VictorLeP Why did you decide to separate different matches?

custom_metrics/privacy.js
VictorLeP: Nope, copy/paste gone wild :)

custom_metrics/privacy.js
VictorLeP: I actually rather expect the frames that contain the ads to provide the transparency metadata. As far as I can tell, the extension traverses all frames (`fetch_metadata_functions.js` in the unpacked extension).

custom_metrics/privacy.js
VictorLeP: I would do both then: `enumerateDevices` sounds more useful for fingerprinting, while `getUserMedia` provides access to the sensitive resource (i.e., being able to capture your webcam feed).

VictorLeP: I might include `getDisplayMedia` as well then.

custom_metrics/privacy.js
VictorLeP: (I'm assuming you're referring to `mediaDevices` and `getUserMedia`)
In general, I match different (sub)strings to balance false negative and false positive matches.
That said, the match on `mediaDevices` is probably not very useful, as `getUserMedia` would capture all (relevant) accesses anyway. I'll leave it out.

VictorLeP: Updated in `a17248f`

custom_metrics/privacy.js
VictorLeP: Updated in a17248f

custom_metrics/privacy.js
max-ostapenko: Maybe you could share how do you plan to query it for the chapter?
I had an idea to put it into some hierarchical form.
GEO API example:
```
geolocation: {
  web_api: {
    navigator_geolocation_getCurrentPosition: findStringInBodies('navigator.geolocation.getCurrentPosition') || findStringInBodies('navigator') && findStringInBodies('geolocation') && findStringInBodies('getCurrentPosition'),
    navigator_geolocation_watchPosition: findStringInBodies('navigator.geolocation.watchPosition') || findStringInBodies('navigator') && findStringInBodies('geolocation') && findStringInBodies('watchPosition'),
  }
}
```
And then we have just to work with booleans in SQL, light and easy.
What do you think?

VictorLeP: Yeah, sounds good - my approach would complicate things too much, as there is no way to consolidate the numbers for the different but related strings.
What would the `web_api` level be useful for though? I don't see why more detail would be needed there.

max-ostapenko: I'm collecting a list of IP geolocation API, like https://ipstack.com/ mentioned in the outline.
But now I rather think to do a PR to: https://github.com/AliasIO/wappalyzer, right?

Is there any other way to get geolocation features?

VictorLeP: My idea was to just search requests for relevant URLs, for example for IPStack: https://api.ipstack.com/.
The only place where that would fail is some sort of 'open-source' IP geolocation library that you could host yourself, but my intuition is that these will be primarily paid services that host their own APIs/libraries.

Wappalyzer could be an alternative approach, but it feels like more work for little gain.

Moreover, the two types of geolocation should be well separated: geolocation APIs are 'hidden' and only use your IP, while `navigator.geolocation` is 'visible' as permission is requested, but the resulting data is much more precise.

VictorLeP: Overall, I agree with refactoring string searches to the or/and forms you proposed, but I don't see a lot of added value in the hierarchy - it seems to me like just another indirection to traverse in the query.

max-ostapenko: Yeah, I totally agree here. Will change variable structure for:
- geolocation,
- mediaDevices,
- globalPrivacyControl,
- permissionsPolicy
- and interestCohort
based on the example then.

max-ostapenko:
@VictorLeP could you add to-do items to the description of this PR?
I've added some that you've completed, but how much is left?

VictorLeP:
> @VictorLeP could you add to-do items to the description of this PR?
> I've added some that you've completed, but how much is left?

I don't seem to be allowed to edit your initial comment (which I presume you mean with 'description')?

VictorLeP:
> > @VictorLeP could you add to-do items to the description of this PR?
> > I've added some that you've completed, but how much is left?
> 
> I don't seem to be allowed to edit your initial comment (which I presume you mean with 'description')?

I'm guessing this is due to the same reason as me not being able to mark the pull request as ready for review: "Only those with write access to this repository can mark a draft pull request as ready for review."

tunetheweb:
You should have full access now @VictorLeP   . Can you try again.

VictorLeP:
Thanks, I don't see an edit option for https://github.com/HTTPArchive/legacy.httparchive.org/pull/211#issue-674074013 yet though. (Just to check I'm looking in the right place: I would expect an 'Edit' item in the menu under the three dots top right of that comment.)

tunetheweb:
Ah looks like this repo is set up differently to the main Web Almanac one so you can’t edit each other’s comments unfortunately.

VictorLeP:
OK, thanks for checking!

VictorLeP:
@max-ostapenko These are the changes to the to-do list I think should be made:

- IAB TCF v1: retrieve consent string -- done in 2770294
- Ads Transparency Spotlight: traverse all frames -- I don't think this is possible? (we can't access 3rd party frames from the custom metrics JS)
- add item for 'privacy policy keywords' -- done (reused from last year)
- add item for 'sensitive resources' with subitems 'permissions policy', 'media devices' and 'geolocation' -- I believe you will still restructure these, then they will be complete?

VictorLeP:
Depending on whether we can use Wappalyzer or not, we might also need custom metrics to detect certain fingerprinting libraries more reliably (e.g., `ClientJS`).

max-ostapenko:
> Depending on whether we can use Wappalyzer or not, we might also need custom metrics to detect certain fingerprinting libraries more reliably (e.g., `ClientJS`).

I thought it would be enough to verify the presence of `ClientJS` object?

VictorLeP:
> > Depending on whether we can use Wappalyzer or not, we might also need custom metrics to detect certain fingerprinting libraries more reliably (e.g., `ClientJS`).
> 
> I thought it would be enough to verify the presence of `ClientJS` object?

Sure, but a custom metric is still needed for that if it's not added to Wappalyzer :)

tunetheweb:
You can open a PR with Wappalyzer if you know how to detect it. And that way everyone gets to benefit from it and not just the Web Almanac.

Here's the PR for FingerprintJs for example: https://github.com/AliasIO/wappalyzer/pull/1790 and a follow up one to improve on that: https://github.com/AliasIO/wappalyzer/pull/3503

rviscomi:
Be aware that if you're relying on the detection being implemented in Wappalyzer, your PR would need to be merged with them before July 1. If the PR is delayed for any reason, the metric may need to be implemented here as a custom metric anyway.

max-ostapenko:
Next release of wappalyser is in about 2 weeks.  @tunetheweb do you know how much time it takes to be used on our crawling servers?

tunetheweb:
We normally sync it up just before the web almanac run to get latest version.

pmeenan:
FWIW, we sync the code to master and don't need to wait for an official release.

VictorLeP:
FYI: PRs https://github.com/AliasIO/wappalyzer/pull/4048 (geolocation) and https://github.com/AliasIO/wappalyzer/pull/4050 (fingerprinting) are pending for adding libraries to Wappalyzer.

custom_metrics/privacy.js
```diff
@@ -15,10 +39,145 @@ return JSON.stringify({
...
+      return ats_meta_tag.content; // JSON message with metadata
+    } else {
+      return null;
+    }
+  })(),
+
+  // FLoC origin trial
+  // Check `meta` tag cf. https://developer.chrome.com/blog/floc/ (complement with relevant header later on)
+  // Test site: https://floc-ot-meta.glitch.me/
+  floc_origin_trial: (() => {
```
max-ostapenko, line 147: @VictorLeP How can we know that a specific trial tag is related to FLoC?
All of the meta tags look similar to me, with just a token being different for each trial/origin.

See here: https://developer.chrome.com/origintrials/#/trials/my

custom_metrics/privacy.js
```diff
@@ -15,10 +46,216 @@ return JSON.stringify({
...
+    return tcData;
+  })(),
+
+  /**
+   * IAB US Privacy User Signal Mechanism “USP API”
+   * https://github.com/InteractiveAdvertisingBureau/USPrivacy
+   *
+   * Test site: https://www.nfl.com/
+   */
+  iab_usp: (() => {
```
max-ostapenko, line 133: @VictorLeP I suggest rewriting tcf detection similar to this version of `iab_usp`.
After some testing, I identified that the previous version of `iab_usp` on nfl.com resulted in false-negative result.

```javascript
(() => {
    let uspData;
    if (typeof __uspapi == 'function') {
      __uspapi('getUSPData', 1, (result, success) => {
        if(success) {
          uspData = result;
        } else {
          uspData = {"error": "Failed to retrieve USPData"};
        }
      });
    } else {
      uspData = null;
    }
    return uspData;
  })()
```
> undefined

as compared to the new one:
```javascript
(() => {
    let uspData = {
      present: typeof window.__uspapi == 'function',
      privacy_string: null,
    };
    if (uspData.present) {
      window.__uspapi('getUSPData', 1, (result, success) => {
        if (success) {
          uspData.privacy_string = result;
        }
      });
    }
    return uspData;
  })()
```
> {present: true, privacy_string: null}

What do you think?

custom_metrics/privacy.js
VictorLeP: You're right, the presence of the token just indicates that the website takes part in an origin trial, not which one specifically. It's possible to extract this from the token:
```js
let tag = "AnQAa+uudTrMiqy0/XRpOyGD9+1P1giQjrAWmPiXNjjzPGt7Tg1pjYGKR4+Do80oKP3ltWfkHhxF3Aq317+8gw0AAABheyJvcmlnaW4iOiJodHRwczovL2Zsb2Mtb3QtbWV0YS5nbGl0Y2gubWU6NDQzIiwiZmVhdHVyZSI6IkludGVyZXN0Q29ob3J0QVBJIiwiZXhwaXJ5IjoxNjI2MjIwNzk5fQ==";
window.atob(content);
> "\u0002t\u0000kë®u:Ì¬´ýti;!÷íOÖ\b°\u0016ø68ó<k{N\riG£Í((ýåµgä\u001e\u001cEÜ\n·×¿¼\r\u0000\u0000\u0000a{\"origin\":\"https://floc-ot-meta.glitch.me:443\",\"feature\":\"InterestCohortAPI\",\"expiry\":1626220799}"
```
The `feature` in the included JSON indicates the type of trial (`InterestCohortAPI` = FLoC trial).

I'm wondering whether the trial token could be useful for any other chapter (other trials might be revelant for them), or whether we should focus on FLoC trial tokens only. Then, we also have the choice to process the token upfront (in a custom metric, for both `meta` tags and headers) or to process them after the fact using BigQuery, to extract the `feature`. What are your thoughts?

custom_metrics/privacy.js
VictorLeP: If it only works correctly in the rewritten version, then of course we should use that one :)
I'll look into rewriting the TCF checks in a similar way.

custom_metrics/privacy.js
VictorLeP: I chatted with a colleague today, and some of the trials might also be useful for at least the Security chapter. I'm thinking of creating a separate custom metrics file (`origin_trial.js`) just for this data point, such that all chapters can use it (without it belonging to the 'privacy' namespace).

custom_metrics/privacy.js
VictorLeP: Refactor done

custom_metrics/privacy.js
max-ostapenko: Nice, following your example I found an example decoding the token:
- https://googlechrome.github.io/OriginTrials/check-token.html

Will remove it from FLoC and add a generic origin-trials metric.

max-ostapenko: Sorry, I haven't reloaded the page and haven't seen your comment about origin_trial.js when writing mine one.
Did you start working on it?

VictorLeP: Not yet, let me know if you prefer to work on it yourself, else I can do it :)

max-ostapenko: Yeah, I've started it yesterday. Will push today.

VictorLeP:
I've added detections of tags related to referrer policies (because these can also be set in HTML) and accesses to the `featurePolicy` property. (these could be considered two additional tasks for the TODO list up top, that have been completed)

max-ostapenko:
@pmeenan There may be multiple 'Origin-Trials' headers with tokens.
How can I parse multiple headers values from $WPT_REQUESTS[0].response_headers which is an object (must have unique keys)?
Please, is it supported in any way?

pmeenan:
@max-ostapenko @rviscomi may know better if any processing has been done, but looking at the [raw HARs](https://www.webpagetest.org/export.php?bodies=1&pretty=1&test=210623_AiDc4F_2d2a9d1a503dfab7c2acbe214084d6da) there are headers in a few places but none are objects indexed by the header name:

response.headers is an array of name/value objects:

Screen Shot 2021-06-25 at 6 06 41 PM

_headers.response is an array of strings with the full name: key header:

Screen Shot 2021-06-25 at 6 07 02 PM

I think the summary tables have headers that were pulled together by the [HA processing code](https://github.com/HTTPArchive/legacy.httparchive.org/blob/3f940b069aefd2b91a21b2e241f54e420f439981/requests.inc#L57). In that case it combines the values of multiple headers with the same key by comma-delimiting them  which is what the [RFC requires](https://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.2) but won't match exactly what came over the wire.

max-ostapenko:
@VictorLeP please could you check the origin-trials parsing and data structure?
I decided to include both meta tag and header options into the custom metric.

max-ostapenko:
@pmeenan did you consider changing the structure of [$WPT_REQUESTS.response_headers](https://docs.webpagetest.org/custom-metrics/#accessing-requests-data-(chrome-only)) similar to HAR's response.headers? (object to array)

pmeenan:
Might be better to see if Rick can create a better summary table from the
HARs. The current requests summary data is from the legacy SQL dumps and
don't have a lot of flexibility.

VictorLeP:
> @VictorLeP please could you check the origin-trials parsing and data structure?
> I decided to include both meta tag and header options into the custom metric.

Looks good to me! A small test suggests that in the case of multiple headers with the same key, the values are collapsed into a comma-separated string in the object, so we don't lose any data there.

VictorLeP:
Depending on https://github.com/HTTPArchive/almanac.httparchive.org/issues/2211, we might still need one more custom metric for `.well-known/gpc.json`, but that shouldn't be too hard. Otherwise, I think everything is done :)

rviscomi:
>Might be better to see if Rick can create a better summary table from the HARs. The current requests summary data is from the legacy SQL dumps and don't have a lot of flexibility.

Yeah this raises a good question about how much work we should be putting into the custom metrics versus post-processed summarization and where to draw the line for the best place for specific types of metrics. 

But to @max-ostapenko's point, it seems like the structure of the headers made available to custom metrics is different from the HAR structure. There's a question of whether we should even do this work in the custom metric (maybe?) but I think parity with the HAR would be good and it'd help with processing repeatable headers, if needed.

@max-ostapenko @VictorLeP is this PR ready for review? Please also add a few test cases to show it working in WPT.

pmeenan:
Argh, sorry - I didn't realize this was in the custom metrics - I thought it was pulling from the summary data.

The standard headers from the dev tools events are in key/value dictionary form: https://chromedevtools.github.io/devtools-protocol/tot/Network/#type-Response

There is headersText which will have the raw headers but that only works for http 1.x responses so it doesn't seem to be worth trying to fall back to.

The custom metrics don't have access to the netlog-processed headers which is what we use for the most accurate data for the HARs so I don't know if we're going to be able to do better for custom metrics.

GJFR:
> Depending on [HTTPArchive/almanac.httparchive.org#2211](https://github.com/HTTPArchive/almanac.httparchive.org/issues/2211), we might still need one more custom metric for `.well-known/gpc.json`, but that shouldn't be too hard. Otherwise, I think everything is done :)

I've made a `well-known.js` as discussed in https://github.com/HTTPArchive/almanac.httparchive.org/issues/2211. You should be able to extend it to check off `.well-known/gpc.json`.

custom_metrics/privacy.js
```diff
@@ -1,12 +1,43 @@
...
...vacybeleid|privacybeleid|privatezza|privatlivspolitik|privatnost|privatnost|privatnosti|privatssphäre|privatumas|privatumo|privatësia|privātuma|privātums|protecció de dades|protecţia datelor|prywatnosci|prywatności|prywatność|regler om fortrolighed|rekisteriseloste|retningslinjer for data|rgpd|sekretess|slapukai|soukromi|soukromí|személyes adatok védelme|súkromie|sīkdatne|sīkdatņu|tietokäytäntö|tietosuoja|tietosuojakäytäntö|tietosuojaseloste|varstvo podatkov|veri i̇lkesi|veri i̇lkesi|veri politikası|vie privée|webbplatsen|yksityisyyden suoja|yksityisyydensuoja|yksityisyys|zasady dotyczące danych|zasebnost|zaštita podataka|zásady ochrany osobných|zásady používání dat|zásady používání dat|zásady využívania údajov|απόρρητο|απόρρητο|πολιτική απορρήτου|πολιτική δεδομένων|προσωπικά δεδομένα|όροι και γνωστοποιήσεις|конфиденциальность|конфіденційність|поверителност|политика за бисквитки|политика за данни|политика использования данных|политика конфиденциальности|политика о подацима|политика о подацима|политика о подацима|политика обработки персональных данных|приватност|приватност|приватност|условия|условия за ползване|מדיניות נתונים|פרטיות|الخصوصية|سياسة البيانات|数据使用政策|數據使用政策|私隱政策|隐私权政策';
```
rviscomi, line 42: Some of these words are pretty small, like `cgu|cgv`. Should you also check to make sure that `\b` boundaries exist on both sides of the words to minimize false positives?

custom_metrics/privacy.js
```diff
@@ -15,10 +46,264 @@ return JSON.stringify({
...
+      present: typeof window.__cmp == 'function',
+      data: null,
+      compliant_setup: null,
+    };
+    // description of `__cmp`: https://github.com/InteractiveAdvertisingBureau/GDPR-Transparency-and-Consent-Framework/blob/master/CMP%20JS%20API%20v1.1%20Final.md#what-api-will-need-to-be-provided-by-the-cmp-
+    if (consentData.present) {
+      // Standard command: 'getVendorConsents'
+      // cf. https://github.com/InteractiveAdvertisingBureau/GDPR-Transparency-and-Consent-Framework/blob/master/CMP%20JS%20API%20v1.1%20Final.md#what-api-will-need-to-be-provided-by-the-cmp-
+      // Test site: ?
+      window.__cmp('getVendorConsents', null, (result, success) => {
```
rviscomi, line 67: Executing application code from the custom metric can get weird.

Does this function have any side-effects on the page? We might want to avoid changing the page, to avoid tainting other metrics.
Does it necessarily support this function signature? We should wrap it in a try/catch in case there are any incompatibilities.

custom_metrics/privacy.js
```diff
@@ -15,10 +46,264 @@ return JSON.stringify({
...
+   *
+   * Test site: https://floc.glitch.me/
+   * Test site: https://www.pokellector.com/
+   *
+   * @todo Check function/variable accesses through string searches (wrappers cannot be used, as the metrics are only collected at the end of the test)
+   */
+  floc: {
+    document_interestCohort:
+      findStringInBodies('document.interestCohort') ||
+      findStringInBodies('interestCohort'),
```
rviscomi, line 175: Are these redundant? Everything matched by the first pattern will also be matched by the second, right?

max-ostapenko:
@GJFR could you create a PR with your `well-known.js`?
We would need to add `parseResponse('/.well-known/gpc.json'),` at least.

custom_metrics/privacy.js
```diff
@@ -15,10 +46,264 @@ return JSON.stringify({
...
+
+  /**
+   * Do Not Track (DNT)
+   * https://www.eff.org/issues/do-not-track
+   *
+   * Test site: https://www.theverge.com/
+   */
+  navigator_doNotTrack:
+    findStringInBodies('navigator.doNotTrack') ||
+    findStringInBodies('doNotTrack'),
```
rviscomi, line 186: Getting slightly concerned about the computational expense of multiple lookups through all of the response bodies. Could this be implemented more efficiently as a regular expression where you extract a regex like `(a|b|c|d)` and map a/b/c/d to the feature detections?

custom_metrics/privacy.js
```diff
@@ -1,12 +1,43 @@
...
+/**
+ * @function findStringInBodies
+ * inspired by https://github.com/HTTPArchive/legacy.httparchive.org/blob/master/custom_metrics/event-names.js
+ *
+ * @param {string} string - String to look for in the response bodies.
+ * @return {boolean} - True, if string was found.
+ */
+function findStringInBodies(string) {
+  try {
+    return response_bodies.some((body) => {
```
rviscomi, line 20: I'd suggest including a filter on the response bodies so that you only check responses that are likely to contain the input strings. For example, only JS and HTML responses.

custom_metrics/privacy.js
```diff
@@ -15,10 +46,264 @@ return JSON.stringify({
...
+    // Test site: https://www.brilio.net/
+    let referrerpolicy_attributes = document.querySelectorAll(
+      '[referrerpolicy]'
+    );
+    // Leave `individual_requests` at `null` if no attributes are found.
+    if (referrerpolicy_attributes.length > 0) {
+    // Build dictionary of occurrences of tag-value pairs.
+      rp.individual_requests = Object.fromEntries(
+        Array.from(referrerpolicy_attributes).map(
+          x => (x.tagName + '|' + x.getAttribute('referrerpolicy'))
```
rviscomi, line 246: WDYT about adding more structure rather than delimiting by `'|'`? For example, you could have an object like `{tagName: x.tagName, referrerpolicy: x.getAttribute('referrerpolicy'), count: 1}` and the `reduce` method could increment the `count` property.

A structure like that is more extensible if needed and may facilitate aggregation/analysis later.

custom_metrics/privacy.js
```diff
@@ -15,10 +46,264 @@ return JSON.stringify({
...
+  })(),
+
+  /**
+   * Media devices
+   * https://developer.mozilla.org/en-US/docs/Web/API/MediaDevices
+   */
+  media_devices: {
+    navigator_mediaDevices_enumerateDevices:
+      findStringInBodies('navigator.mediaDevices.enumerateDevices') ||
+      findStringInBodies('navigator') ||
```
rviscomi, line 278: Will this return true for any response that includes the word `navigator`?

max-ostapenko:
> Please also add a few test cases to show it working in WPT.
@rviscomi We've added test websites in the comments, is that what you mean?

@pmeenan ok, got it. My take was to write one custom metric that will check all optional feature signs: in meta tags or in headers. I know learned that for the most accurate analysis we have to combine it with SQL over headers in the requests table.

custom_metrics/privacy.js
max-ostapenko: Yes, you're right. Will change it to make sure all of the keywords are part of the specific JS file.

custom_metrics/privacy.js
max-ostapenko: Yeap. It is a typo.

VictorLeP:
> > Please also add a few test cases to show it working in WPT.
> 
> @rviscomi We've added test websites in the comments, is that what you mean?

I think he means provide links to actual tests on https://www.webpagetest.org/, I'll try to get to those today.

> @pmeenan ok, got it. My take was to write one custom metric that will check all optional feature signs: in meta tags or in headers. I know learned that for the most accurate analysis we have to combine it with SQL over headers in the requests table.

It's indeed a bit annoying that we cannot do the whole extraction upfront to have clean data already in the requests table, but must merge the meta tag and header values in the SQL query. (Given that the literal definition of `<meta http-equiv="...">` is to simulate HTTP headers, I could imagine that there are more cases where the same issue arises.)
@max-ostapenko I'm thinking we should then revert to just storing the raw base64 value for the origin value meta tags in the custom metric, and then doing the extraction of all the attributes in the SQL query? We can use a JS UDF there, so it would just be a matter of moving over the function statement.

custom_metrics/privacy.js
VictorLeP: It's the only way to obtain the necessary data. Since they're (marked as) pure getters (in the documentation), I wouldn't expect any side effects.

I'll get to the try/catch, even though any library implementing this API is supposedly [required to support these methods](https://github.com/InteractiveAdvertisingBureau/GDPR-Transparency-and-Consent-Framework/blob/master/TCFv2/IAB%20Tech%20Lab%20-%20CMP%20API%20v2.md#how-does-the-cmp-provide-the-api).

VictorLeP: done in 74b46c6

custom_metrics/privacy.js
VictorLeP: [`\b` doesn't seem to work well with non-ASCII characters](https://stackoverflow.com/questions/2449779/why-cant-i-use-accented-characters-next-to-a-word-boundary/2449877#2449877), of which we have a few in the regex, though I'm not sure we really need them. Otherwise, it makes sense.

VictorLeP: Already enforced word boundaries in 905142e

custom_metrics/privacy.js
VictorLeP: I implemented an approach similar to yours in cee66cb, I don't know if that is what you had in mind.
(I wanted to store it as an object originally as well, but didn't think further than to use a hash map, which fails because objects with the same properties are not equal.)

custom_metrics/privacy.js
VictorLeP: Filtered on `Document` and `Script` types in ebd97c8. 
`type` looks to me like the only property that allows us to determine whether a response is JS/HTML, or am I missing something?
Could `XHR` and `Fetch` also be sources of relevant HTML/scripts?

custom_metrics/privacy.js
max-ostapenko: I'm on it

custom_metrics/privacy.js
max-ostapenko: I think `Document` and `Script` are fine.

custom_metrics/privacy.js
max-ostapenko: done in 3cb05be19658a80de7603ce75c1eed3c04cec357

custom_metrics/privacy.js
max-ostapenko: done in 3cb05be

max-ostapenko:
> @max-ostapenko I'm thinking we should then revert to just storing the raw base64 value for the origin value meta tags in the custom metric, and then doing the extraction of all the attributes in the SQL query? We can use a JS UDF there, so it would just be a matter of moving over the function statement.

@VictorLeP let's keep it the way it is now for crawling. But I'll add a base64 string to the custom metric.
Then we'll be able to verify if there is any value to go the hard way (compared to HAR data) and will adjust it to UDF.
What do you think?

VictorLeP:
> > @max-ostapenko I'm thinking we should then revert to just storing the raw base64 value for the origin value meta tags in the custom metric, and then doing the extraction of all the attributes in the SQL query? We can use a JS UDF there, so it would just be a matter of moving over the function statement.
> 
> @VictorLeP let's keep it the way it is now for crawling. But I'll add a base64 string to the custom metric.
> Then we'll be able to verify if there is any value to go the hard way (compared to HAR data) and will adjust it to UDF.
> What do you think?

Sounds excellent, it'll be interesting to compare the two sources and see if there is any meaningful difference.

VictorLeP:
WebPageTest runs:

- [brilio.net](https://www.webpagetest.org/custom_metrics.php?test=210629_BiDcGX_85903298411c85a3094bc1535752b9ab&run=1&cached=0)
- [cnet.com](https://www.webpagetest.org/custom_metrics.php?test=210629_BiDcCE_658f3447a24ef5fbf18b3f4fe6df8455&run=1&cached=0)
- [nfl.com](https://www.webpagetest.org/custom_metrics.php?test=210629_BiDc7P_742323b7f7df83e277b36ab398d70a68&run=1&cached=0)
- [pokellector.com](https://www.webpagetest.org/custom_metrics.php?test=210629_AiDc33_1def85166676b8f9c9e86e0d6d861add&run=1&cached=0)
- [rtl.de](https://www.webpagetest.org/custom_metrics.php?test=210629_BiDcRF_a7ff29e4fb700c8890d6a149c1e06ff8&run=1&cached=0)
- [global-privacy-control.glitch.me](https://www.webpagetest.org/custom_metrics.php?test=210629_AiDcRC_484cdfc7779cfb98aeabc478f08d88a5&run=1&cached=0)
- [floc.glitch.me](https://www.webpagetest.org/custom_metrics.php?test=210629_BiDcWN_f500ba0e9f7c07003d1f2a433cd2750c&run=1&cached=0)
- [floc-ot-meta.glitch.me](https://www.webpagetest.org/custom_metrics.php?test=210629_AiDcG9_bf224e59b6655a4ed57dc3e42e3ee6ec&run=1&cached=0)
- [floc-ot-header.glitch.me](https://www.webpagetest.org/custom_metrics.php?test=210629_BiDcP5_0af0a85d3d0133b736322e56bbba58f3&run=1&cached=0)

custom_metrics/origin-trials.js
```diff
@@ -0,0 +1,169 @@
...
+
+let unique_tokens = tokens.filter(
+  (value, index, self) => self.indexOf(value) === index
+);
+
+let origin_trials = [];
+
+unique_tokens.forEach((token) => {
+  let origin_trial = validate(token);
+  origin_trial.token = token;
```
max-ostapenko, line 164: @VictorLeP I found it's already logged [here](https://github.com/HTTPArchive/legacy.httparchive.org/blob/master/custom_metrics/almanac.js#L221)
Can we remove it?

VictorLeP: Ah, sure, no use in duplicating the data. (I'm assuming it won't be too terribly difficult to extract the relevant nodes using SQL.)

max-ostapenko: Those metatags don't cover tokens from headers, so I'll leave this one.

P.S. I think we will have a hard time decoding token values with SQL: `TextDecoder` and `atob` API are not available in UDF. So there will be an effort to make polyfills work 😑

max-ostapenko:
@VictorLeP @rviscomi seems we are ready for merge

custom_metrics/origin-trials.js
```diff
@@ -0,0 +1,156 @@
...
+    obj = JSON.parse(json);
+  } catch (e) {
+    console.error(e);
+    validityElem = 'Invalid JSON';
+    return;
+  }
+
+  originElem = obj.origin;
+  subdomainElem = obj.isSubdomain ? 'Yes' : 'No';
+  thirdpartyElem = obj.isThirdParty ? 'Yes' : 'No';
```
rviscomi, line 88: Seems like it'd be more useful to leave these as booleans, rather than having to do a string comparison later during aggregation/analysis.
```suggestion
  subdomainElem = !!obj.isSubdomain;
  thirdpartyElem = !!obj.isThirdParty;
```

custom_metrics/origin-trials.js
```diff
@@ -0,0 +1,156 @@
...
+
+  const utf8Decoder = new TextDecoder('utf-8', {fatal: true});
+
+  // Base64-decode the token into a Uint8Array.
+  let tokenStr;
+  try {
+    tokenStr = atob(tokenElem);
+  } catch (e) {
+    console.error(e);
+    validityElem = 'Invalid Base64';
```
rviscomi, line 35: What's the purpose of assigning to `validityElem` before returning? Doesn't seem like it's ever used.

Maybe you should initialize `origin_trial_metadata` earlier, update the validity property as needed, and return the entire object when there are processing errors. That way it'll still include the validity data but none of the other processed data.

custom_metrics/origin-trials.js
```diff
@@ -0,0 +1,156 @@
...
+let tokens = meta_tags.map(tag => tag.content);
+
+requests.forEach(request => {
+  let header = getParameterCaseInsensitive(request.response_headers, 'origin-trial');
+  if (header) {
+    tokens = tokens.concat(header.replaceAll(' ', '').split(','));
+  }
+});
+
+let unique_tokens = tokens.filter((value, index, self) => self.indexOf(value) === index);
```
rviscomi, line 147: Is this the same thing?
```suggestion
let unique_tokens = Array.from(new Set(tokens));
```

custom_metrics/privacy.js
```diff
@@ -1,24 +1,312 @@
...
+ * inspired by https://github.com/HTTPArchive/legacy.httparchive.org/blob/master/custom_metrics/event-names.js
+ *
+ * @param {string} pattern - Regex pattern to match in the response bodies.
+ * @return {boolean} - True, if pattern was matched.
+ */
+function testPropertyStringInResponseBodies(pattern) {
+  try {
+    let re = new RegExp(pattern);
+    return response_bodies
+      .filter(body => body.type === 'Document' || body.type === 'Script')
```
rviscomi, line 25: I'd suggest moving this filtering out of the function so that it's only done once as opposed to each time the function is called. (response bodies are very large so these optimizations can be significant, especially after running them on 8M+ pages)

custom_metrics/privacy.js
```diff
@@ -1,24 +1,312 @@
 //[privacy]
 // Uncomment the previous line for testing on webpagetest.org
 
+// README! Instructions for adding a new custom metric for the Web Almanac.
+// 2. If the value requires more than one line of code, evaluate it in an IIFE, eg `(() => { ... })()`. See `link-nodes`.
+// 3. Test your change by following the instructions at https://github.com/HTTPArchive/almanac.httparchive.org/issues/33#issuecomment-502288773.
+// 4. Submit a PR to update this file.
+
+const response_bodies = $WPT_BODIES;
```
rviscomi, line 9:
```suggestion
const response_bodies = $WPT_BODIES.filter(body => body.type === 'Document' || body.type === 'Script');
```

custom_metrics/privacy.js
```diff
@@ -1,24 +1,312 @@
...
-    var pattern = new RegExp(words, 'i');
+    var pattern = new RegExp('\\b(?:' + words + ')\\b', 'i');
+
+    return Array.from(document.querySelectorAll('a')).filter(a => a.innerText.match(pattern))
+      .length;
```
rviscomi, line 52: Would it be useful to output which privacy wording was used? It may facilitate debugging later or provide interesting data to analyze about which words are more popular.

custom_metrics/privacy.js
```diff
@@ -1,24 +1,312 @@
...
+
+  /**
+   * FLoC
+   *
+   * Test site: https://floc.glitch.me/
+   * Test site: https://www.pokellector.com/
+   *
+   * @todo Check function/variable accesses through string searches (wrappers cannot be used, as the metrics are only collected at the end of the test)
+   */
+  document_interestCohort: testPropertyStringInResponseBodies('document.+interestCohort'),
```
rviscomi, line 186: Not sure about this pattern. `.+` could include anything so `document` and `interestCohort` could be very far apart in the script. Also, do you mean to use the literal `.` character here, since it's a regex?

I would expect that you'd only want to match instances of `document.interestCohort` and not things like `documentFOOinterestCohort`.

max-ostapenko: The point is that we can't know for sure if there will be a call using a global name or some local variable ([example](https://github.com/fingerprintjs/fingerprintjs/blob/f61e348c741b5b904e1cdc6265098a51bff6a9b2/src/utils/browser.ts))


So this multiple keyword match should be better than just `testPropertyStringInResponseBodies('interestCohort')`.
And no, there is no literal `.`.

custom_metrics/origin-trials.js
```diff
@@ -0,0 +1,152 @@
...
+  versionElem = '' + version;
+  if (version !== 2 && version !== 3) {
+    origin_trial_metadata.validityElem = 'Unknown version';
+    return origin_trial_metadata;
+  }
+
+  // Pull the fields out of the token.
+  if (token.length < 69) {
+    validityElem = 'Token is too short';
+    return;
```
rviscomi, line 55:
```suggestion
    origin_trial_metadata.validityElem = 'Token is too short';
    return origin_trial_metadata;
```

custom_metrics/privacy.js
```diff
@@ -1,24 +1,312 @@
...
-    var pattern = new RegExp(words, 'i');
+    let pattern = new RegExp('\\b(?:' + words + ')\\b', 'i');
 
-    return Array.from(document.querySelectorAll('a')).filter((a) =>
-      a.innerText.match(pattern)
-    ).length;
+    let privacy_links = Array.from(document.querySelectorAll('a')).filter(a => a.innerText.match(pattern));
+
+    return privacy_links.map(link => link.innerText);
```
rviscomi, line 52: Returning the whole text may make aggregation difficult later, for example accounting for additional text / whitespace / case-sensitivity. WDYT about mapping links to the matching word pattern specifically?

rviscomi (commented):
Last two comments

custom_metrics/privacy.js
VictorLeP: Or both? Having the whole text might be more useful for debugging, the specific pattern more for additional analysis.

VictorLeP: Stored both in 00f7160. Feel free to revert, but in that case we could also consider using [`test`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/test) instead of `match`.
WPT run (rtl.de): https://www.webpagetest.org/custom_metrics.php?test=210630_AiDcD7_c350b1535df7dec7bddbafff01698b79&run=1&cached=0

rviscomi (approved):

PR #213, opened 2021-06-23
JavaScript 2021 custom metrics

NishuGoel:
Progress on https://github.com/HTTPArchive/almanac.httparchive.org/issues/2141

adds a custom metrics to capture the number of AJAX requests

[WebPageTest result](https://webpagetest.org/custom_metrics.php?test=210629_AiDc83_dfeae5f4d889d7d68f7263535bb84371&run=1&cached=0)

### DIFF
--- a/custom_metrics/javascript.js
+++ b/custom_metrics/javascript.js
@@ -40,8 +40,24 @@ return JSON.stringify({
             return null;
         }
     })(),
-    // Returns the content-types of requests.
-    content_types: requests.map((req) => req.response_headers['content-type']),
+
+    web_component_specs: (() => {
+
+        var getNodeName = (el) => el.nodeName.toLowerCase()
+        var elements_with_hyphen = Array.from(document.getElementsByTagName("*")).filter(e => e.nodeName.includes('-'));
+        var unique_elements_with_hyphen = [...new Set(elements_with_hyphen.map(e => e.nodeName))].map(element => { return elements_with_hyphen.find(e => e.nodeName === element) });
+        var web_components = unique_elements_with_hyphen.filter(e => customElements.get(e.nodeName.toLowerCase()));
+
+        // Checking which web comp spec is used by found web_components (template, shadow_dom, custom_elements)
+        var shadow_roots = web_components.filter(e => e.shadowRoot)
+        var template = web_components.filter(e => Array.from(e.children).some(r => r.nodeName.toLowerCase() === 'template'))
+        var web_component_specs = {
+            'custom_elements': web_components.map(getNodeName),
+            'shadow_roots': shadow_roots.map(getNodeName),
+            'template': template.map(getNodeName)
+        }
+        return web_component_specs
+    }),
 
     script_tags: (() => {
         let script_tags = Array.from(document.querySelectorAll('script'));
@@ -57,6 +73,10 @@ return JSON.stringify({
             referrerpolicy: script_tags.filter(tag => tag.referrerpolicy).length,
             src: script_tags.filter(tag => tag.src).length,
             type_module: script_tags.filter(tag => tag.type == 'module').length,
+            //deprecated attribute adoption
+            charset: script_tags.filter(tag => tag.charset).length,
+            language: script_tags.filter(tag => tag.attributes['language']).length,
+
         };
 
         return script_data;
--- a/custom_metrics/javascript.js
+++ b/custom_metrics/javascript.js
@@ -73,6 +73,8 @@ return JSON.stringify({
             referrerpolicy: script_tags.filter(tag => tag.referrerpolicy).length,
             src: script_tags.filter(tag => tag.src).length,
             type_module: script_tags.filter(tag => tag.type == 'module').length,
+            async_and_defer: script_tags.filter(tag => tag.async && tag.defer).length,
+            defer_without_src: script_tags.filter(tag => tag.defer && !tag.src).length,
             //deprecated attribute adoption
             charset: script_tags.filter(tag => tag.charset).length,
             language: script_tags.filter(tag => tag.attributes['language']).length,
--- a/custom_metrics/javascript.js
+++ b/custom_metrics/javascript.js
@@ -1,0 +1,46 @@
+return JSON.stringify({
+    'num-ajax-requests': (() => {
+        // Returns the number of ajax requests per page.
+        var ajax_apis = ['xmlhttprequest', 'fetch', 'beacon'];
+        return window.performance.getEntriesByType('resource').filter(r => ajax_apis.includes(r.initiatorType)).length;
+
+    })(),
+    // Returns the number of iframes per page.
+    'iframe': document.getElementsByTagName('iframe').length,
+    'ajax_protocol': (() => {
+        // Returns the percentage of http 2 protocol used.
+        try {
+            var requests = $WPT_REQUESTS;
+            var usage_percent
+            if (requests.length > 0) {
+                var num_requests = requests.length;
+                var http2_count = requests.filter(r => r.protocol === "h2").length;
+                usage_percent = (http2_count / num_requests) * 100;
+            }
+            else {
+                usage_percent = null;
+            }
+
+            return usage_percent;
+        } catch (e) {
+            return null;
+        }
+    })(),
+    'content-types': (() => {
+        // Returns the content-types of requests.
+        try {
+            var requests = $WPT_REQUESTS;
+            if (requests.length > 0) {
+                return requests.map(req => req.request_headers["accept"])
+            }
+            else {
+                return null
+            }
+
+        } catch (e) {
+            return null;
+        }
+    })(),
+    // Returns the number of ajax requests using beacon.
+    'beacon_ajax_usage': window.performance.getEntriesByType('resource').filter(r => r.initiatorType === 'beacon').length
+});
--- a/custom_metrics/javascript.js
+++ b/custom_metrics/javascript.js
@@ -72,6 +72,7 @@ return JSON.stringify({
             nonce: script_tags.filter(tag => tag.nonce).length,
             referrerpolicy: script_tags.filter(tag => tag.referrerpolicy).length,
             src: script_tags.filter(tag => tag.src).length,
+            inline: script_tags.filter(tag => !tag.src).length,
             type_module: script_tags.filter(tag => tag.type == 'module').length,
             async_and_defer: script_tags.filter(tag => tag.async && tag.defer).length,
             defer_without_src: script_tags.filter(tag => tag.defer && !tag.src).length,
--- a/custom_metrics/javascript.js
+++ b/custom_metrics/javascript.js
@@ -40,23 +40,6 @@ return JSON.stringify({
             return null;
         }
     })(),
-
-    webcomponents_shadowDOM: (() => {
-        // Returns the number of web components or shadow roots on a web page.
-        try {
-            var elements_with_hyphen = Array.from(document.getElementsByTagName("*")).filter(e => e.nodeName.includes('-'));
-            var shadow_roots = elements_with_hyphen.filter(e => e.shadowRoot)
-            var web_components = elements_with_hyphen.filter(e => customElements.get(e.nodeName.toLowerCase()));
-            let element_data = {
-                total_potential_web_components: elements_with_hyphen,
-                web_components: web_components,
-                shadow_roots: shadow_roots
-            }
-            return element_data;
-        } catch (e) {
-            return null;
-        }
-    })(),
     // Returns the content-types of requests.
     content_types: requests.map((req) => req.response_headers['content-type']),
 
--- a/custom_metrics/javascript.js
+++ b/custom_metrics/javascript.js
@@ -54,12 +54,26 @@ return JSON.stringify({
         try {
             var requests = $WPT_REQUESTS;
             if (requests.length > 0) {
-                var types = requests.map(req => req.response_headers["Content-Type"])
+                return requests.map(req => req.response_headers['content-type']);
             }
             else {
-                return null
+                return requests.length
             }
 
+        } catch (e) {
+            return null;
+        }
+    })(),
+    'memory_usage': (() => {
+        /*  Returns the memory usage of a page - JS and dom elements only
+            Important - webpagetest needs flags to be test: 
+            --disable-web-security, --no-site-isolation */
+        try {
+            return new Promise((resolve) => {
+                performance.measureUserAgentSpecificMemory().then((value) => {
+                    resolve(JSON.stringify(value));
+                });
+            });
         } catch (e) {
             return null;
         }
--- a/custom_metrics/javascript.js
+++ b/custom_metrics/javascript.js
@@ -5,23 +5,46 @@ return JSON.stringify({
         return window.performance.getEntriesByType('resource').filter(r => ajax_apis.includes(r.initiatorType)).length;
 
     })(),
+    // Returns the number of ajax requests using beacon.
+    'beacon_ajax_usage': window.performance.getEntriesByType('resource').filter(r => r.initiatorType === 'beacon').length,
     // Returns the number of iframes per page.
     'iframe': document.getElementsByTagName('iframe').length,
-    'ajax_protocol': (() => {
+    'requests_protocol': (() => {
         // Returns the percentage of http 2 protocol used.
         try {
             var requests = $WPT_REQUESTS;
-            var usage_percent
             if (requests.length > 0) {
-                var num_requests = requests.length;
-                var http2_count = requests.filter(r => r.protocol === "h2").length;
-                usage_percent = (http2_count / num_requests) * 100;
+                var ajaxs = window.performance.getEntriesByType('resource').filter(r => ['xmlhttprequest', 'fetch', 'beacon'].includes(r.initiatorType))
+                var resources = window.performance.getEntriesByType('resource').filter(r => r.initiatorType !== 'xmlhttprequest' && r.initiatorType !== 'fetch' && r.initiatorType !== 'beacon')
+                return JSON.stringify({
+                    'total_requests': requests.length,
+                    'total_h1': requests.filter(r => ["http/1.1", "http/1"].includes(r.protocol)).length,
+                    'ajax_h1': ajaxs.filter(r => ["http/1.1", "http/1"].includes(r.nextHopProtocol)).length,
+                    'resources_h1': resources.filter(r => ["http/1.1", "http/1"].includes(r.nextHopProtocol)).length,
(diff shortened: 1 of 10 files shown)


max-ostapenko:
Hi @NishuGoel, Looks good!

Let's include all the chapter's custom metrics in this PR.
I see there are some more to be added, so maybe we start with `javascript.js` to have a clear chapter relation.
And if there is anything that may be shared - we can put it to separate files.

rviscomi:
+1 to @max-ostapenko's suggestion to use `javascript.js`

NishuGoel:
@max-ostapenko That is all the custom metrics we need, think we can go ahead with this.

custom_metrics/javascript.js
```diff
@@ -0,0 +1,81 @@
...
+
+    })(),
+    // Returns the number of ajax requests using beacon.
+    'beacon_ajax_usage': window.performance.getEntriesByType('resource').filter(r => r.initiatorType === 'beacon').length,
+    // Returns the number of iframes per page.
+    'iframe': document.getElementsByTagName('iframe').length,
+    'requests_protocol': (() => {
+        // Returns the percentage of http 2 protocol used.
+        try {
+            var requests = $WPT_REQUESTS;
```
rviscomi, line 15: Create `const requests = $WPT_REQUESTS;` at the top of the file so it can be reused across any individual metrics that need it. WPT will expand all instances of `$WPT_REQUESTS` (which can be very large) so it's good to minimize the number of times it needs to do that.

custom_metrics/javascript.js
```diff
@@ -0,0 +1,81 @@
...
+                var resources = window.performance.getEntriesByType('resource').filter(r => r.initiatorType !== 'xmlhttprequest' && r.initiatorType !== 'fetch' && r.initiatorType !== 'beacon')
+                return JSON.stringify({
+                    'total_requests': requests.length,
+                    'total_h1': requests.filter(r => ["http/1.1", "http/1"].includes(r.protocol)).length,
+                    'ajax_h1': ajaxs.filter(r => ["http/1.1", "http/1"].includes(r.nextHopProtocol)).length,
+                    'resources_h1': resources.filter(r => ["http/1.1", "http/1"].includes(r.nextHopProtocol)).length,
+                    'total_h2': requests.filter(r => ["h2", "http/2", "http2"].includes(r.protocol)).length,
+                    'ajax_h2': ajaxs.filter(r => ["h2", "http/2", "http2"].includes(r.nextHopProtocol)).length,
+                    'resources_h2': resources.filter(r => ["h2", "http/2", "http2"].includes(r.nextHopProtocol)).length,
+                    'total_h3': requests.filter(r => r.protocol === "http/3").length,
```
rviscomi, line 27: There are a few ways H3 can be represented. Would we need to account for those here too?

Context: https://twitter.com/tunetheweb/status/1408026535408381956 cc @tunetheweb 

Are there any other reasons why the sum of each of these metrics wouldn't add up to `total_requests`?

custom_metrics/javascript.js
```diff
@@ -0,0 +1,81 @@
...
+        } catch (e) {
+            return null;
+        }
+    })(),
+    'num_webcomp_shadow': (() => {
+        // Returns the number of web components or shadow roots on a web page.
+        try {
+            var elements_with_hyphen = Array.from(document.body.getElementsByTagName("*")).filter(e => e.nodeName.includes('-'));
+            return JSON.stringify({
+                'total_potential_WCs': elements_with_hyphen.length,
```
rviscomi, line 44: Use full name for clarity

```suggestion
                'total_potential_web_components': elements_with_hyphen.length,
```

custom_metrics/javascript.js
```diff
@@ -0,0 +1,81 @@
...
+                return null;
+            }
+        } catch (e) {
+            return null;
+        }
+    })(),
+    'num_webcomp_shadow': (() => {
+        // Returns the number of web components or shadow roots on a web page.
+        try {
+            var elements_with_hyphen = Array.from(document.body.getElementsByTagName("*")).filter(e => e.nodeName.includes('-'));
```
rviscomi, line 42: Would it also be useful to expose the list of distinct element names?

rviscomi, line 42: Do you also need to enforce that the hyphen is in the middle of the node name, or can it be at the start/end?

custom_metrics/javascript.js
```diff
@@ -0,0 +1,81 @@
...
+    'content-types': (() => {
+        // Returns the content-types of requests.
+        try {
+            var requests = $WPT_REQUESTS;
+            if (requests.length > 0) {
+                return requests.map(req => req.response_headers['content-type']);
+            }
+            else {
+                return requests.length
+            }
```
rviscomi, line 61: I think we can assume `requests` will always be > 0. Even if it's an empty array, `map` would return an empty array.
```suggestion
            return requests.map(req => req.response_headers['content-type']);
```

rviscomi (changes requested):

custom_metrics/num_ajax_requests.js
```diff
@@ -0,0 +1,2 @@
+var ajax_apis = ['xmlhttprequest', 'fetch', 'beacon']
+return window.performance.getEntriesByType('resource').filter(r => ajax_apis.includes(r.initiatorType)).length
```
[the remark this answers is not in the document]
rviscomi: Ping

custom_metrics/javascript.js
```diff
@@ -0,0 +1,81 @@
+return JSON.stringify({
+    'num-ajax-requests': (() => {
```
max-ostapenko, line 2: @NishuGoel should we maybe do it with SQL:
```sql
SELECT
  AVG(requests_cnt) AS avj_num_ajax_requests
FROM (
  SELECT
    page,
    COUNT() AS requests_cnt
  FROM
    `httparchive.latest.requests_desktop`
  WHERE
    JSON_QUERY(payload, "$._initiator_type") IN ('xmlhttprequest', 'fetch', 'beacon')
  GROUP BY
    page )
```

custom_metrics/javascript.js
```diff
@@ -0,0 +1,81 @@
...
+            return JSON.stringify({
+                'total_potential_WCs': elements_with_hyphen.length,
+                'custom_elements': elements_with_hyphen.filter(e => customElements.get(e.nodeName.toLowerCase())).length,
+                'shadow_roots': elements_with_hyphen.filter(e => e.shadowRoot).length,
+            })
+        } catch (e) {
+            return null;
+        }
+    })(),
+    'content-types': (() => {
```
max-ostapenko, line 52: @NishuGoel would this SQL give the similar result?
```sql
SELECT
    pageid,
    REGEXP_EXTRACT(reqOtherHeaders, 'Content-Type = (.+?),'),
    COUNT(1)
FROM `httparchive.summary_requests.2021_06_01_desktop`
GROUP BY 1,2
ORDER BY 3 DESC
LIMIT 10
```

custom_metrics/javascript.js
pankajparkar: Good point, just curious do you know any framework or libraries that do it? So I can handle work on handling this usecase.

custom_metrics/javascript.js
pankajparkar: @max-ostapenko Yes, this should return desired results, we may need to change the RegEx a bit.

custom_metrics/javascript.js
```diff
@@ -0,0 +1,81 @@
...
+            }
+            else {
+                return requests.length
+            }
+
+        } catch (e) {
+            return null;
+        }
+    })(),
+    'memory_usage': (() => {
```
pankajparkar: @NishuGoel 
Since we can not disable the flags, shall we remove this custom metrics?

custom_metrics/javascript.js
```diff
@@ -0,0 +1,107 @@
...
+          resolve(JSON.stringify(value));
+        });
+      });
+    } catch (e) {
+      return null;
+    }
+  })(),
+
+  script_properties: (() => {
+    let script_tags = Array.from(document.querySelectorAll('script'));
```
pankajparkar, line 90: Add try catch

custom_metrics/javascript.js
```diff
@@ -0,0 +1,107 @@
+//[javascript]
+// Uncomment the previous line for testing on webpagetest.org
+
+// Please, refer to instructions for adding a custom metric in almanac.js.
+
+const requests = $WPT_REQUESTS;
+
+return JSON.stringify({
+  'num-ajax-requests': (() => {
+    // Returns the number of ajax requests per page.
```
pankajparkar, line 10: Add try catch

pankajparkar (changes requested):

custom_metrics/javascript.js
max-ostapenko: Done in 28ea84f1a9655545a140aba5f35ec15f083bfe60

custom_metrics/javascript.js
pankajparkar: @NishuGoel Please use below to find distinct elements

```
var elements_with_hyphen = Array.from(document.body.getElementsByTagName("*")).filter(e => e.nodeName.includes('-'));
var unique_web_components = [...new Set( elements_with_hyphen.map( e => e.nodeName))].map( element=> { return elements_with_hyphen.find(e => e.nodeName === element) } );
```

custom_metrics/javascript.js
max-ostapenko: Why?

custom_metrics/javascript.js
max-ostapenko: There is a `has_shadow_root.js` from last year.
Let's update it with these additional details?

NishuGoel: Thanks @pankajparkar
Removing this metric as it does not cover anything more than as covered in `has_shadow_root.js`

Done in [f6a6e24](https://github.com/HTTPArchive/legacy.httparchive.org/pull/213/commits/f6a6e24231692b19551b0e317620088b3fafd4ee)

custom_metrics/num_ajax_requests.js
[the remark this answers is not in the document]
NishuGoel: we planned to get the api (xmlhttprequest, fetch) difference using usage_counters earlier, but good to have all in one place.

done in [472eed7](https://github.com/HTTPArchive/legacy.httparchive.org/pull/213/commits/472eed75d0f7255fae2582ef78af3be88a25f9dd)

max-ostapenko:
@NishuGoel I don't have permission to resolve comments in your PR.
Please close whatever you think doesn't need attention anymore, or let me know what needs my attention.

custom_metrics/javascript.js
NishuGoel: added the other representations. Updated the metric here [a6d3ea9](https://github.com/HTTPArchive/legacy.httparchive.org/pull/213/commits/a6d3ea97fb9286033a63d184ebdb2558520a6f9a)

custom_metrics/javascript.js
NishuGoel: covered already in `has_shadow_root.js`
Done in [f6a6e24](https://github.com/HTTPArchive/legacy.httparchive.org/pull/213/commits/f6a6e24231692b19551b0e317620088b3fafd4ee)

custom_metrics/javascript.js
NishuGoel: Done in [472eed7](https://github.com/HTTPArchive/legacy.httparchive.org/pull/213/commits/472eed75d0f7255fae2582ef78af3be88a25f9dd)

custom_metrics/num_ajax_requests.js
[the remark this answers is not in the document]
max-ostapenko: Can this be achieved with SQL [as here](https://github.com/HTTPArchive/legacy.httparchive.org/pull/213#discussion_r660613086) or is it different?

custom_metrics/javascript.js
NishuGoel: @max-ostapenko yup, good to go with, 
the custom metric gives us the output as : `{xmlhttprequest: 0, fetch: 15, beacon: 6, total: 21}`

might need to tweak the SQL or just go with the metric. Have kept it in the code anyway in [a6d3ea9](https://github.com/HTTPArchive/legacy.httparchive.org/pull/213/commits/a6d3ea97fb9286033a63d184ebdb2558520a6f9a):

max-ostapenko: I found out that this initiator is of [a different nature](https://developer.chrome.com/docs/devtools/network/reference/#requests), so yes - we need these custom metrics to identify AJAX requests API.

custom_metrics/javascript.js
```diff
@@ -0,0 +1,73 @@
...
+    // Returns the content-types of requests.
+    content_types: requests.map((req) => req.response_headers['content-type']),
+
+    script_tags: (() => {
+        let script_tags = Array.from(document.querySelectorAll('script'));
+
+        let script_data = {
+            total: script_tags.length,
+            async: script_tags.filter(tag => tag.async).length,
+            defer: script_tags.filter(tag => tag.defer).length,
```
rviscomi, line 52: One thing the 2020 chapter looked at was the usage of both [async and defer](https://github.com/HTTPArchive/almanac.httparchive.org/blob/main/sql/2020/javascript/async_and_defer.sql) on the same script. Would be good to support that in the custom metric.

Also given the [special note](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script#attr-defer) for `defer` on MDN about the attribute not being used if the `src` attribute is absent, it may be nice to include a metric for scripts that use that antipattern.

```suggestion
            async: script_tags.filter(tag => tag.async).length,
            defer: script_tags.filter(tag => tag.defer).length,
            async_and_defer: script_tags.filter(tag => tag.async && tag.defer).length,
            defer_without_src: script_tags.filter(tag => tag.defer && !tag.src).length,
```

custom_metrics/javascript.js
```diff
@@ -0,0 +1,73 @@
...
+            crossorigin: script_tags.filter(tag => tag.crossorigin).length,
+            integrity: script_tags.filter(tag => tag.integrity).length,
+            nomodule: script_tags.filter(tag => tag.nomodule).length,
+            nonce: script_tags.filter(tag => tag.nonce).length,
+            referrerpolicy: script_tags.filter(tag => tag.referrerpolicy).length,
+            src: script_tags.filter(tag => tag.src).length,
```
rviscomi, line 58: Is it true that `total - src` is the number of inline script blocks? Would it be useful to calculate that directly?
```suggestion
            src: script_tags.filter(tag => tag.src).length,
            inline: script_tags.filter(tag => !tag.src).length,
```

custom_metrics/javascript.js
```diff
@@ -0,0 +1,73 @@
...
+            return null;
+        }
+    })(),
+    // Returns the content-types of requests.
+    content_types: requests.map((req) => req.response_headers['content-type']),
+
+    script_tags: (() => {
+        let script_tags = Array.from(document.querySelectorAll('script'));
+
+        let script_data = {
```
rviscomi, line 49: For any of these, are you interested in extracting the attribute values? For example, browsers may treat the script element as a data block depending on the [type](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script#attr-type) attribute.

rviscomi, line 49: Would you also be interested to track adoption of the [deprecated attributes](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script#deprecated_attributes) `charset` and `language`?

rviscomi (changes requested):

custom_metrics/javascript.js
NishuGoel: getting rid of content_type metric as this updated SQL serves what we need.

custom_metrics/javascript.js
NishuGoel: @max-ostapenko Could you please answer this?

custom_metrics/javascript.js
NishuGoel: Isn't this already covered as a metric due to the 2020 chapter? Do we need to rewrite this? @rviscomi

custom_metrics/javascript.js
NishuGoel: Captured [here](https://github.com/HTTPArchive/legacy.httparchive.org/pull/213/commits/70a30783d235d94dd48866b0c1839f110565e106)

NishuGoel: Not sure, so not capturing any attribute values

custom_metrics/javascript.js
NishuGoel: added these anyway

rviscomi: It was covered last year but the problem was that the custom metric didn't exist and we (I was the analyst) implemented it by parsing the HTML in SQL, which is expensive and error-prone. It'd be much better to process the data in JS instead.

custom_metrics/javascript.js
rviscomi: SGTM

rviscomi: LGTM

custom_metrics/javascript.js
NishuGoel: Added this anyway

NishuGoel:
Hi @rviscomi @max-ostapenko All covered in latest changes, removed metrics that could be done through SQL.
Please have a look and merge if all well.

cc: @pankajparkar

custom_metrics/javascript.js
```diff
@@ -0,0 +1,96 @@
...
+            nonce: script_tags.filter(tag => tag.nonce).length,
+            referrerpolicy: script_tags.filter(tag => tag.referrerpolicy).length,
+            src: script_tags.filter(tag => tag.src).length,
+            inline: script_tags.filter(tag => !tag.src).length,
+            type_module: script_tags.filter(tag => tag.type == 'module').length,
+            async_and_defer: script_tags.filter(tag => tag.async && tag.defer).length,
+            defer_without_src: script_tags.filter(tag => tag.defer && !tag.src).length,
+            //deprecated attribute adoption
+            charset: script_tags.filter(tag => tag.charset).length,
+            language: script_tags.filter(tag => tag.attributes['language']).length,
```
rviscomi, line 81: A bit more semantic (could be done for all of these attribute metrics, but seems like `language` is a special case)
```suggestion
            language: script_tags.filter(tag => tag.hasAttribute('language')).length,
```

rviscomi (approved):
Just one nit, then this is good to merge.

custom_metrics/javascript.js
NishuGoel: there we go!

PR #216, opened 2021-06-27
New A11Y metrics

obto:
Progress on https://github.com/HTTPArchive/almanac.httparchive.org/issues/2147

This PR adds five new metrics:

1. `form_control_a11y_tree`: A slimmed down version of the browser accessibility tree for form controls (select, input, textarea, buttons)
2. `fieldset_radio_checkbox`: Documents how many radios and checkboxes there are. Then checks how many of those inputs are correctly placed in a fieldset with a legend
3. `required_form_controls`: Scans for all the required form controls on the page. Looks for `required`, `aria-required`, or an asterisk at the start or end of the visible label
4. Track `<audio>` elements and their `<track>`s
5. Addition of a `total_with_track` metric to the `videos` section in almanac.js

This code has been tested on custom built pages ([example](https://output.jsbin.com/ziwubinatu/1)) and amazon.com (stress test for large amounts of elements) and passed the tests.

[Amazon.com test](https://www.webpagetest.org/custom_metrics.php?test=210627_AiDc7T_04ad325e19e194b517da98f2e38bc2cb&run=1&cached=0)

[Custom page test](https://www.webpagetest.org/result/210627_BiDc2P_3c5bf0159172753a230ea9e6895f0fd8/)

### DIFF
--- a/custom_metrics/a11y.js
+++ b/custom_metrics/a11y.js
@@ -210,4 +210,172 @@ return JSON.stringify({
   screen_reader_classes: captureAndLogError(() => {
     return document.querySelectorAll('.sr-only, .visually-hidden').length > 0;
   }),
+  form_control_a11y_tree: captureAndLogError(() => {
+    function doesMatchAny(string, regexps) {
+      let found = false;
+      for (const regexp of regexps) {
+        if (regexp.test(string)) {
+          return true;
+        }
+      }
+
+      return false;
+    }
+
+    const attributes_to_track = [
+      /^aria-.+$/,
+      /^type$/,
+      /^id$/,
+      /^name$/,
+      /^placeholder$/,
+      /^accept$/,
+      /^autocomplete$/,
+      /^autofocus$/,
+      /^capture$/,
+      /^max$/,
+      /^maxlength$/,
+      /^min$/,
+      /^minlength$/,
+      /^required$/,
+      /^readonly$/,
+      /^pattern$/,
+      /^multiple$/,
+      /^step$/,
+    ];
+    function addControlToStats(node, accumulator) {
+      const control_stats = {
+        type: node.node_info.nodeType.toLowerCase(),
+        attributes: {},
+        properties: {},
+        accessible_name: node.name.value || '',
+        accessible_name_sources: [],
+        role: node.role.value || '',
+      };
+
+      // Store all attribute information
+      for (let [key, value] of Object.entries(node.node_info.attributes || {})) {
+        key = key.toLowerCase();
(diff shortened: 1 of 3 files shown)


custom_metrics/a11y.js
```diff
@@ -210,4 +210,184 @@ return JSON.stringify({
...
+      for (const regexp of regexps) {
+        if (regexp.test(string)) {
+          return true;
+        }
+      }
+
+      return false;
+    }
+
+    const attributes_to_track = [
```
rviscomi, line 225: Might be simpler to bundle all of these in a single regex that can be tested at once, rather than requiring iteration. For example: `/^(aria-.+|type|id|...)$/i`. Then `doesMatchAny` can be replaced with `attributes_to_track.test(key)`. I've also added the `i` flag to handle case insensitivity rather than lowercasing the key manually.

custom_metrics/a11y.js
```diff
@@ -210,4 +210,184 @@ return JSON.stringify({
...
+
+    const allowed_control_types = [
+      'input',
+      'select',
+      'textarea',
+      'button',
+    ];
+
+    const stats_of_controls = [];
+    for (const node of $WPT_ACCESSIBILITY_TREE) {
```
rviscomi, line 298: I'd suggest declaring a const at the top of the file that aliases `$WPT_ACCESSIBILITY_TREE`, so that there's only one version of it that needs to be expanded in case anyone wants to add a new custom metric that depends on it.

custom_metrics/a11y.js
```diff
@@ -210,4 +210,184 @@ return JSON.stringify({
...
+      const label = getVisibleLabel(element);
+      if (!label) {
+        return false;
+      }
+
+      if (label.substr(0, 1) === '*' || label.substr(-1, 1) === '*') {
+        return true;
+      }
+
+      return false;
```
rviscomi, line 369:
```suggestion
      return label.startsWith('*') || label.endsWith('*');
```

rviscomi (changes requested):

custom_metrics/a11y.js
obto: Always forget about the "new" methods. Even though they've been out for a while now

custom_metrics/a11y.js
obto: Good idea. Made this change in the latest commit

custom_metrics/a11y.js
obto: Change made in latest commit

rviscomi (approved):

PR #217, opened 2021-06-27
New mobile web metric

obto:
Progress on https://github.com/HTTPArchive/almanac.httparchive.org/issues/2151

This PR adds one new metric: `link_protocols_used`.

It's purpose is to list all the different types of protocols used in links on the page (sms, tel, https, ...)

It has been tested on a [custom page](https://output.jsbin.com/vafomucutu/1) and amazon which has a slew of links. Everything works as expected.

[Amazon.com](https://www.webpagetest.org/custom_metrics.php?test=210627_AiDcA0_21f55b07202f0919c68f5da4a8afcfff&run=1&cached=0)

[Custom page](https://www.webpagetest.org/custom_metrics.php?test=210627_BiDcNH_445e26032bcf071fe5eff777839660b5&run=1&cached=0)

### DIFF
--- a/custom_metrics/almanac.js
+++ b/custom_metrics/almanac.js
@@ -424,6 +424,33 @@ return JSON.stringify({
 
     return inputNodes;
   })(),
+  'link_protocols_used': (() => {
+    const protocols_and_count = {};
+    const anchors = document.querySelectorAll('a[href]');
+    for (const anchor of anchors) {
+      try {
+        const url = new URL(anchor.href);
+
+        // Get the protocol and remove the :
+        const protocol = (url.protocol || '').toLowerCase().slice(0, -1);
+        if (!protocol) {
+          continue;
+        }
+
+        // Increase the counter
+        if (protocols_and_count[protocol]) {
+          protocols_and_count[protocol]++;
+        } else {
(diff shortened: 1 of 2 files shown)


custom_metrics/almanac.js
```diff
@@ -424,6 +424,33 @@ return JSON.stringify({
 
     return inputNodes;
   })(),
+  'link_protocols_used': (() => {
+    const protocols_and_count = {};
+    const anchors = document.querySelectorAll('a[href]');
+    for (const anchor of anchors) {
+      try {
+        const url = new URL(anchor.href);
```
rviscomi, line 432: Are you only interested in absolute URLs? Or for example if the a[href] value is `/foo` do you want to consider that as having a protocol of `https` if the base page is also HTTPS?
```suggestion
        const url = new URL(anchor.href, location.href);
```

obto: The `href` property on anchor elements already does this for you. You always get the fully qualified url

rviscomi: WOW that's cool :)

rviscomi (approved):

PR #218, opened 2021-06-28
Structured Data 2021

GregBrimble:
https://github.com/HTTPArchive/almanac.httparchive.org/issues/2174
https://docs.google.com/document/d/19KDSv4olAXUHUV6hq4X4Cb-lNziqvVesgXXxVktrw4c/edit#

### DIFF
--- a/custom_metrics/structured-data.js
+++ b/custom_metrics/structured-data.js
@@ -85,10 +85,12 @@ try {
       try {
         function dumpTagAttributes(tags) {
           return tags.map((tag) => {
-            const attributes = tag.getAttributeNames();
             const json_tag = {};
-            attributes.map((attribute) => {
-              json_tag[attribute] = tag.getAttribute(attribute);
+            ["rel", "href", "name", "property", "content"].map((attribute) => {
+              const value = tag.getAttribute(attribute);
+              if (value) {
+                json_tag[attribute] = tag.getAttribute(attribute);
+              }
             });
             return json_tag;
           });
@@ -103,7 +105,6 @@ try {
         function gatherStructuredData(d) {
           let target = {
             jsonld_scripts: [],
-            raw_html: "",
             present: {
               json_ld: false,
               microdata: false,
@@ -134,9 +135,6 @@ try {
             ...d.querySelectorAll('script[type="application/ld+json"i]'),
           ].map((script) => script.innerHTML);
           target.present.json_ld = target.jsonld_scripts.length > 0;
-
-          // Microdata, RDFa & Microformats
-          target.raw_html = d.documentElement.outerHTML;
 
           // Microdata
           target.present.microdata =
@@ -217,19 +215,19 @@ try {
 
           // Twitter
           target.twitter = dumpTagAttributes([
-            ...d.querySelectorAll('[name^="twitter:" i]'),
+            ...d.querySelectorAll('meta[name^="twitter:" i]'),
           ]);
           target.present.twitter = target.twitter.length > 0;
 
           // Facebook
           target.facebook = dumpTagAttributes([
-            ...d.querySelectorAll('[property^="fb:" i]'),
+            ...d.querySelectorAll('meta[property^="fb:" i]'),
           ]);
           target.present.facebook = target.facebook.length > 0;
 
           // OpenGraph
           target.opengraph = dumpTagAttributes([
-            ...d.querySelectorAll('[property^="og:" i]'),
+            ...d.querySelectorAll('meta[property^="og:" i]'),
           ]);
           target.present.opengraph = target.opengraph.length > 0;
 
--- a/custom_metrics/structured-data.js
+++ b/custom_metrics/structured-data.js
@@ -86,12 +86,13 @@ try {
         function dumpTagAttributes(tags) {
           return tags.map((tag) => {
             const json_tag = {};
-            ["rel", "href", "name", "property", "content"].map((attribute) => {
-              const value = tag.getAttribute(attribute);
-              if (value) {
-                json_tag[attribute] = tag.getAttribute(attribute);
+            ["rel", "href", "name", "property", "content"].forEach(
+              (attribute) => {
+                if (tag.hasAttribute(attribute)) {
+                  json_tag[attribute] = tag.getAttribute(attribute);
+                }
               }
-            });
+            );
             return json_tag;
           });
         }
@@ -191,7 +192,6 @@ try {
             "hReview",
             "hReview-aggregate",
             "adr",
-            "geo",
           ].forEach((name) => {
             let items = d.querySelectorAll('[class~="' + name + '" i]');
             if (items.length > 0) {
@@ -202,6 +202,17 @@ try {
               });
             }
           });
+
+          const geoItems = d.querySelectorAll(
+            '[class~="geo" i] [class~="latitude" i]'
+          );
+          if (geoItems.length > 0) {
+            target.present.microformats_classic = true;
+            target.microformats_classic_types.push({
+              name: "geo",
+              count: geoItems.length,
+            });
+          }
 
           // Dublin Core, Twitter, Facebook & OpenGraph
           target.meta_tags = dumpTagAttributes([...d.querySelectorAll("meta")]);
--- a/custom_metrics/structured-data.js
+++ b/custom_metrics/structured-data.js
@@ -1,0 +1,144 @@
+//[structured_data]
+// Uncomment the previous line for testing on webpagetest.org
+
+var _logs = [];
+// saves the error details in the results log property.
+// returns the same error object so that it can be also used as the return value for a property.
+function logError(context, messageOrException, exception = null) {
+  let error = { type: "error", context: context };
+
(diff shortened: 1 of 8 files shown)


custom_metrics/structured-data.js
```diff
@@ -0,0 +1,144 @@
...
+
+          return target;
+        }
+
+        let result = {};
+        result.rendered = gatherStructuredData(document);
+
+        let rawHtmlDocument = getRawHtmlDocument();
+
+        if (rawHtmlDocument) {
```
rviscomi, line 117: IIRC the SEO chapter differentiates between raw/rendered markup to get a better sense of dynamically added content. Is that as interesting to the Structured Data chapter, or could this be simplified to look at only the rendered markup?

rviscomi (commented):
I know this is in Draft mode so I won't critique the code too closely, but I wanted to ask a high-level question about the choice of evaluating the raw HTML.

Please also keep in mind that we have two days to merge this PR, so mark it as "ready for review" ASAP.

custom_metrics/structured-data.js
vdwijngaert: Good point. It might be interesting to report on the way SD is added (dynamic vs static). As the SEO chapter is doing it, might be something worth looking into?

GregBrimble: I think we'll just leave it in, if that's okay. We'll take a look once we've got the data, and if there's interesting stuff (e.g. some SD being overwritten by injection, or inconsistencies being introduced by injection), then we'll be able to talk about it. Otherwise, we'll just stick with the final rendered page, I imagine.

jono-alderson: Yep! I don't think this will form a key part of the storytelling, but agreed that it's worth reviewing for any interesting nuances.

GregBrimble: We've had a bit of a discussion about the Microdata, RDFa & Microformat HTML stuff here: https://httparchive.slack.com/archives/C021GGN9W4D/p1624916290080500.

Again, I think if it's okay, we'll just leave the HTML in, but I'm also just away to add a simple "flag if found" so we can report on `% of pages which include that type of SD`.

We might play around with a HTML parser later, but attempting it in these custom metrics is likely too brittle.

GregBrimble:
The only non-JSON-LD type I haven't enhanced in this most recent commit is Dublin Core. Dublin Core can be represented in a number of different ways, so again, I think if we tried here, we'd likely fail in collecting some implementations. Perhaps more can be done later with the raw HTML.

jono-alderson:
> The only non-JSON-LD type I haven't enhanced in this most recent commit is Dublin Core. Dublin Core can be represented in a number of different ways, so again, I think if we tried here, we'd likely fail in collecting some implementations. Perhaps more can be done later with the raw HTML.

I'd be happy limiting this to extracting any `<meta` tag with a `name` property beginning with `DC` (case-insensitive). I think that give us the majority of 'normal' usage.

GregBrimble:
Cool. As long as you say that disclaimer in your writing, that's good with me :) I'll add that in now.

rviscomi (changes requested):

custom_metrics/structured-data.js
rviscomi: I tested this custom metric on the NYTimes website and it produces 2.6 MB of data, in the `raw_html` alone. Keep in mind that we store all of these results in a BigQuery table for millions of pages, so this would become prohibitively large very quickly.

Also note that we already have the raw HTML (see the `response_bodies` dataset) so outputting it here defeats the purpose of using custom metrics, which is to make use of JS APIs at runtime to more efficiently process the document. In short, if we're parsing strings of HTML, we're doing it wrong 😅

custom_metrics/structured-data.js
```diff
@@ -0,0 +1,271 @@
...
+              });
+            }
+          });
+
+          // Dublin Core, Twitter, Facebook & OpenGraph
+          target.meta_tags = dumpTagAttributes([...d.querySelectorAll("meta")]);
+          target.link_tags = dumpTagAttributes([...d.querySelectorAll("link")]);
+
+          // Dublin Core
+          target.dublin_core = dumpTagAttributes([
```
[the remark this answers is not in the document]
GregBrimble: We're now just storing the following properties: `["rel", "href", "name", "property", "content"]`

custom_metrics/structured-data.js
```diff
@@ -0,0 +1,271 @@
...
+              target.present.microformats_classic = true;
+              target.microformats_classic_types.push({
+                name: name,
+                count: items.length,
+              });
+            }
+          });
+
+          // Dublin Core, Twitter, Facebook & OpenGraph
+          target.meta_tags = dumpTagAttributes([...d.querySelectorAll("meta")]);
```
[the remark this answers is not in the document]
GregBrimble: We do definitely want to have the `content` values for these tags, so I think we should still do our own thing separate to what's in `almanac.js`, but we've restricted our storage to just the following attributes now: `["rel", "href", "name", "property", "content"]`.

custom_metrics/structured-data.js
GregBrimble: Removed the `raw_html` part. I think this chapter is focusing on JSON-LD, and we'll be able to talk about general usage of non-JSON-LD with what we have already, without the `raw_html` bit. So we've removed that to save on storage :)

custom_metrics/structured-data.js
```diff
@@ -0,0 +1,271 @@
...
+          };
+
+          // JSON-LD
+          target.jsonld_scripts = [
+            ...d.querySelectorAll('script[type="application/ld+json"i]'),
+          ].map((script) => script.innerHTML);
+          target.present.json_ld = target.jsonld_scripts.length > 0;
+
+          // Microdata, RDFa & Microformats
+          target.raw_html = d.documentElement.outerHTML;
```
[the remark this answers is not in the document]
GregBrimble: Done!

custom_metrics/structured-data.js
```diff
@@ -0,0 +1,269 @@
...
+
+  _custom_metrics = {
+    structured_data: (() => {
+      try {
+        function dumpTagAttributes(tags) {
+          return tags.map((tag) => {
+            const json_tag = {};
+            ["rel", "href", "name", "property", "content"].map((attribute) => {
+              const value = tag.getAttribute(attribute);
+              if (value) {
```
rviscomi, line 91: Should this handle cases where the attribute name exists, but either the value is omitted (`<foo bar>`) or the value is empty (`<foo bar="">`)?
```suggestion
              if (tag.hasAttribute(attribute)) {
```

custom_metrics/structured-data.js
```diff
@@ -0,0 +1,269 @@
...
+    return _rawHtmlDocument;
+  }
+
+  _custom_metrics = {
+    structured_data: (() => {
+      try {
+        function dumpTagAttributes(tags) {
+          return tags.map((tag) => {
+            const json_tag = {};
+            ["rel", "href", "name", "property", "content"].map((attribute) => {
```
rviscomi, line 89: Nit: not actually mapping the results
```suggestion
            ["rel", "href", "name", "property", "content"].forEach((attribute) => {
```

custom_metrics/structured-data.js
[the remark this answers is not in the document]
rviscomi: I raised the question because it's not immediately clear how all meta/link elements are related to structured data. There would need to be an additional post-processing step to analyze the attributes and determine which ones are relevant to the chapter and aggregate/summarize them in a useful way. Usually that's done at the same time as the custom metric processing to simplify the SQL workload.

Is your intention to emit more data than you need so that you can analyze it later? If so this would be similar to my feedback on the raw HTML, that we should process the data now.

custom_metrics/structured-data.js
```diff
@@ -0,0 +1,269 @@
...
+            "hCard",
+            "hListing",
+            "hMedia",
+            "hProduct",
+            "hRecipe",
+            "hResume",
+            "hReview",
+            "hReview-aggregate",
+            "adr",
+            "geo",
```
rviscomi, line 194: These may lead to a lot of false positives. Is there any way to be more confident that `class="geo"` uses microformats?

GregBrimble:
The full meta/link tags were saved before we have more specific checks later. Twitter, Facebook and OpenGraph should all be covered now. Due to its expressiveness, it's just Dublin Core that might be missing some of the capturing that we've implemented.

If it's a storage space concern, I think we could nix it (@jono-alderson has already said they're fine with just capturing tags which begin with `DC`). But if we _can_ keep it, we might find other Dublin Core tags, which we'd otherwise lose.

jono-alderson:
Yeah, let's simplify and accept that we might miss some stuff. First year, let's keep it simple.

GregBrimble:
Consider it gone!

rviscomi (approved):
Thanks for helping get this over the finish line, excited to see what insights we can unlock using this new data!

PR #219, opened 2021-06-28
Security 2021 custom metrics

GJFR:
Also includes renaming and extending ecommerce custom metric for well-known URLs as per https://github.com/HTTPArchive/almanac.httparchive.org/issues/2211.


rviscomi:
Is this ready for review?

max-ostapenko:
@GJFR Could you please add, as ,emtioned in #211 
```javascript
// privacy
parseResponse('/.well-known/gpc.json', r => {
  return r.text().then(text => {
    let data = {
      'gpc': null
    };
    let gpc_data = JSON.parse(text);
    if (typeof gpc_data.gpc == 'boolean') {
      data.gpc = gpc_data.gpc;
    }
    return data;
  });
}),
```

custom_metrics/well-known.js
```diff
@@ -0,0 +1,87 @@
...
+          data["expires"] = line.substring(9);
+        } else if (line.startsWith("Policy: ")) {
+          data["policy"] = line.substring(8);
+        }
+      }
+      return data;
+    });
+  })
+]).then((all_data) => {
+  return JSON.stringify(all_data);
```
rviscomi, line 84: This returns an array of objects. WDYT about returning an object whose keys are the URLs? That may make analysis a bit easier, for example to extract `$["well-known"]["/.well-known/security.txt"]` as opposed to `$["well-known"].find(obj => obj.path == "/.well-known/security.txt")`.

`parseResponse` could return an array like `[url, resultObj]` and here we'd return `JSON.stringify(Object.fromEntries(all_data))`.

custom_metrics/well-known.js
```diff
@@ -0,0 +1,87 @@
...
+function parseResponse(url, parser) {
+  return fetchWithTimeout(url).then(request => {
+    let result = {'path': url};
+    if(!request.redirected && request.status === 200) {
+      result['found'] = true;
+      if(parser) {
+        let promise = parser(request);
+        if (promise) {
+          return promise.then(data => {
+            result["data"] = data;
```
rviscomi, line 26: Nit: Mixing single/double quotes styles in this file. Pick one for consistency.

custom_metrics/well-known.js
```diff
@@ -0,0 +1,87 @@
...
+    });
+  }),
+  parseResponse('/.well-known/security.txt', r => {
+    return r.text().then(text => {
+      let data = {};
+      if (text.startsWith("-----BEGIN PGP SIGNED MESSAGE-----")) {
+        data["signed"] = true;
+      } else {
+        data["signed"] = false;
+      }
```
rviscomi, line 68: I think this will immediately overwrite `signed = true`. What about something like this instead?

```suggestion
      let data = {
        signed: false
      };
      if (text.startsWith("-----BEGIN PGP SIGNED MESSAGE-----")) {
        data["signed"] = true;
      }
```

rviscomi (changes requested):

GJFR:
@max-ostapenko Your code has been added 👍 

@rviscomi Thank you for your comments! I've updated the code. I'm gonna do a quick check and tweak the robots.txt data collecting. Will mark as ready for review ASAP.

custom_metrics/well-known.js
```diff
@@ -0,0 +1,102 @@
...
+      return data;
+    });
+  }),
+  parseResponse('/.well-known/security.txt', r => {
+    return r.text().then(text => {
+      let data = {
+        'signed': false
+      };
+      if (text.startsWith('-----BEGIN PGP SIGNED MESSAGE-----')) {
```
SaptakS, line 81: ❤️

GJFR:
WPT test runs:

- [Test google.com](https://www.webpagetest.org/custom_metrics.php?test=210629_AiDcRB_14635572c03d17579853bedd9fdc2308&run=1&cached=0)
- [Test facebook.com](https://www.webpagetest.org/custom_metrics.php?test=210629_BiDc3R_667ab8efa4b8627a9f20b882b84241b2&run=1&cached=0)
- [Test github.com](https://www.webpagetest.org/custom_metrics.php?test=210629_AiDcD0_2a5013351e7654f9d8c9e21acea4f787&run=1&cached=0)

SaptakS:
WPT tests look good to me.

custom_metrics/well-known.js
```diff
@@ -0,0 +1,104 @@
...
+      let data = {'disallows': {}};
+      let currUserAgent = null;
+      for(let line of text.split('\n')) {
+        if (line.startsWith('User-agent: ')) {
+          currUserAgent = line.substring(12);
+          if (data['disallows'][currUserAgent] === undefined) {
+            data['disallows'][currUserAgent] = [];
+          }
+        } else if (line.startsWith('Disallow: ')) {
+          data['disallows'][currUserAgent].push(line.substring(10));
```
rviscomi, line 72: Is this logging too much info? Curious what you plan to do with it. If you're only interested in high-level stats like whether disallows are used at all and for which UAs, or even how many paths are disallowed, is there a way to summarize those stats in the results instead of outputting every path?

SaptakS: We would be interested to check how (and if) the login endpoints are disallowed. A regex to check some keywords like "login", "sign-in", and only capture those paths might be ok. @tomvangoethem @nrllh thoughts?

GJFR: Just did a quick check on a dozen of sites. I think we will cover most login pages if we filter the URL on `login`, `log-in`, `signin`, and `sign-in`.

Just let me know if you'd like to have the code updated this way and I'll push the change @SaptakS @tomvangoethem @nrllh.

PS: I also came across a false positive: https://github.com/account-login refers to an account, but is included in `robots.txt` weirdly enough. 🤔

GJFR: Oh, I completely forgot a discussion I had with Tom. We should also filter on `admin`, might be interesting as well.

VictorLeP: Maybe also `auth`, `sso` and `account`?
Were you planning on matching 1) the whole path, 2) any path segment, or 3) any substring of the path?

GJFR: > Maybe also `auth`, `sso` and `account`?

Could also be interesting!

> Were you planning on matching 1) the whole path, 2) any path segment, or 3) any substring of the path?

I was thinking any substring of the URL (not only path). There's also signin.aws.amazon.com for instance.

VictorLeP: > > Were you planning on matching 1) the whole path, 2) any path segment, or 3) any substring of the path?
> 
> I was thinking any substring of the URL (not only path). There's also signin.aws.amazon.com for instance.

That implies expanding every path in `robots.txt` to its full path and matching on that? What I'm getting at: [subdomains appear to require their own `robots.txt`](https://developers.google.com/search/docs/advanced/robots/robots_txt). That would mean that, to determine that `signin.aws.amazon.com` cannot be crawled because of `robots.txt`, you'd have to find a wildcard entry `*` in `signin.aws.amazon.com/robots.txt`. That goes beyond just parsing the contents of `robots.txt` and matching on that. (And technically, _any_ entry of `robots.txt` for a domain which contains a keyword will result in a match.)

SaptakS: Yeah. Subdomains might be a little bit trickier to get, I agree. But at the least we should do any substring of the path.

GJFR: Hmm, good point... I wasn't aware that subdomains need their own `robots.txt`.

Still, I think there is no need for fetching `robots.txt` from subdomains that contain certain keywords since the idea is for an attacker to probe the local `robots.txt` to collect potentially sensitive endpoints. So just matching any substring of the path will be sufficient for the sec chapter.

VictorLeP: Indeed, the 'sensitive' subdomains will need to be discovered through other means anyway (subdomain enumeration etc.), so that is a distinct security issue.

GJFR:
I filtered on all keywords discussed in this thread.

A few thoughts:
- If we are not that interested in `User-agent` entries in itself, we could remove reported user-agents that do not have any matched `Disallow` paths to reduce clutter.
- As shown in the test results below: `author` seems a popular false positive for `auth`. I'm sure we will encounter similar false positives in the crawl data. In my opinion, we should make the filter more strict in the querying stage and not on the custom metric level for two reasons:
  - We could overtighten the filter and miss out on endpoints. It's safer to try out more strict filters when we have the crawl data.
  - It's more practical to put as less filtering logic at custom metric level as possible. For example, if later this metric will be used in a longitudinal analysis spanning multiple crawls, it will be more difficult to get it right if the custom metric filter has been adjusted between crawls. Not that familiar with the database/querying costs, though, so can't say whether it will be worth the (limited?) additional cost.

WPT test runs:
- [google.com](https://www.webpagetest.org/custom_metrics.php?test=210630_AiDc0A_3da1cf004a037f6318a087d2f48d1ce6&run=1&cached=0)
- [facebook.com](https://www.webpagetest.org/custom_metrics.php?test=210630_AiDcD2_8c70a5445873bc641ceb6c455ffd5d86&run=1&cached=0)
- [github.com](https://www.webpagetest.org/custom_metrics.php?test=210630_BiDc6S_4257f1dd0e2819c4a356daee2b1037e4&run=1&cached=0)
- [ieee.org](https://www.webpagetest.org/custom_metrics.php?test=210630_BiDcBH_8bfb393ec74a8e54908689428cea77f2&run=1&cached=0)

custom_metrics/well-known.js
tomvangoethem: I'm not sure I understand the security sensitiveness/impact of adding the login/authentication path to `robots.txt`, especially if we're only checking for paths containing known strings. Unless I'm missing something, the only security issue I see is if someone would hide an admin panel behind a random path and put that path in the `robots.txt`.

GJFR: As far as I understand, `robots.txt` could be abused from an automated attacker's perspective. While attackers can probe server responses for variations of `/login`, `/admin`, etc., this would be inefficient. Furthermore, websites could use more unique endpoints for these purposes (that perhaps still contain the `login`/`admin`/... substring).

So in that sense, matching `robots.txt` for these substrings could aid an automated attack to find potentially sensitive endpoints more efficiently, and could identify more unique variations.

That's just my interpretation though, maybe this metric was proposed with another vulnerability in mind.

rviscomi (approved):

rviscomi:
LGTM thanks everyone!

PR #220, opened 2021-06-29
Responsive images custom metrics

eeeps:
Progress on https://github.com/HTTPArchive/almanac.httparchive.org/issues/2144

This is not quite ready to be merged...

It all works, but I _need_ to:
- think about keys on the image entries in order to join them with other sources of info on these Elements. @rviscomi if you have any ideas here I am all ears.
- make sure I get or can get the bits/pixel, by getting the loaded resource bytesize and attatching it to this and the approximate dimensions.

Then there are some nice-to-haves... but that's the must do list.

PR'ing to get some more eyes on what's here now and comments in a place other than twitter.

My "test suite" such as it is is: https://codepen.io/eeeps/debug/Exmxwzq
Here's a successful WPT run against that: https://webpagetest.org/custom_metrics.php?test=210629_AiDc5N_8d945980cad843b7ed362df6d53ef17e&run=1&cached=0

### DIFF
--- a/custom_metrics/responsive_images.js
+++ b/custom_metrics/responsive_images.js
@@ -1,4 +1,4 @@
-//[responsive-images]
+// [responsive-images]
 // Uncomment the previous line for testing on webpagetest.org
 
 return JSON.stringify({
@@ -308,6 +308,25 @@ function getImgData( img ) {
 	imgData.url = img.currentSrc || img.src;
 	imgData.totalCandidates = totalNumberOfCandidates( img );
 	
+	if (imgData.hasHeight) {
+		imgData.heightAttribute = img.getAttribute('height');
+	}
+	if (imgData.hasWidth) {
+		imgData.widthAttribute = img.getAttribute('width');
+	}
+	if (imgData.hasAlt) {
+		imgData.altAttribute = img.getAttribute('alt');
+	}
+	if (imgData.hasCustomDataAttributes) {
+		imgData.customDataAttributes = Object.keys( img.dataset );
+	}
+	
+	// duplicates metrics captured in images.js, but these could ease additional analysis in combination with everything else we're capturing...
+	imgData.clientWidth = parseInt( img.clientWidth );
+	imgData.clientHeight = parseInt( img.clientHeight );
+	imgData.naturalWidth = parseInt( img.naturalWidth );
+	imgData.naturalHeight = parseInt( img.naturalHeight );
+	
 	if ( imgData.isInPicture ) {
 		const pictureFeatures_ = pictureFeatures( img );
 		imgData.pictureMediaSwitching = pictureFeatures_.mediaSwitching;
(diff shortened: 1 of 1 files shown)


custom_metrics/responsive_images.js
```diff
@@ -0,0 +1,895 @@
...
+
+}
+
+// the candidate <source> elements of a given <picture>
+// takes a <picture> Element
+// returns a plain ol' Array of <source> Elements
+function pictureSources( picture ) {
+	// all <source>s that come BEFORE the <img>
+	// https://twitter.com/zcorpan/status/1409778636153032706
+	return [ ...picture.querySelectorAll( 'source:not( img ~ source )' ) ];
```
zcorpan, line 78: Maybe this is overly pedantic, but this can still return incorrect `source` elements (e.g. grandchildren).

I would pass in the `img` as argument, and walk to the `parentNode` then through its `children` until you find the `img` that was passed in.

custom_metrics/responsive_images.js
```diff
@@ -0,0 +1,895 @@
...
+// we check for srcset and sizes use separately, in order to consider values supplied by the parent <picture> / <source> siblings
+// TODO consider loading (although there's already a metric for this?), decode,  crossorigin, referrerpolicy
+function imgFeatures( img ) {
+
+	return {
+		hasSrc: img.hasAttribute( 'src' ),
+		hasAlt: img.hasAttribute( 'alt' ),
+		isInPicture: img.parentNode.tagName === "PICTURE",
+		hasCustomDataAttributes: Object.keys( img.dataset ).length > 0,
+		hasWidthAndHeight: img.hasAttribute( 'width' ) &&
```
nhoizey, line 229: Would it be useful to know if/how many images have either `width` or `height` but not both?

custom_metrics/responsive_images.js
```diff
@@ -0,0 +1,959 @@
+[responsive-images]
```
rviscomi, line 1: Comment this out when committing. Uncomment locally for testing.
```suggestion
//[responsive-images]
```

rviscomi (approved):
This was a lot to review but I mentally stepped through most of the custom metric code and it generally LGTM. The test case shows it working as expected. This is ready to merge from my perspective.

It's good to have other reviewers to look at the specific implementation details, so I'll wait for one more LGTM before merging, unless we run out of time.

custom_metrics/responsive_images.js
eeeps: Done

custom_metrics/responsive_images.js
eeeps: Done!

eeeps:
@rviscomi thank you so very, very much.

Since you reviewed...
- Addressed all review comments from @zcorpan (here and on Twitter) and @nhoizey
- Added the current loaded resource URL as a key to join to things
- Added a few significant new metrics (!) in 
af3fbef and ab392ee. Which: I'm sorry. But bits/pixel in particular was something I was very excited to add to the analysis this year (and needed all of this density-awareness in order to calculate).

I'll see if I can get another full review by EOD.

zcorpan:
The resource selection logic LGTM.

eeeps:
Added a few more values to the reported JSON -- all stuff we were either already looking at in order to calculate other things, or the values of attributes we were previously only testing for the presense of.

There is some duplication with what's captured in images.js, but I think having those things here will make analysis that would have required a join less error-prone.

---

@rviscomi asked for some WPT results; here they are:

- [harvard.edu](https://webpagetest.org/custom_metrics.php?test=210630_AiDc2H_fceff5223b76124a01a01a9f317edb61&run=1&cached=0) (contains a few srcsets with w descriptors)
- [nytimes.com](https://webpagetest.org/custom_metrics.php?test=210630_BiDcYZ_3a0f6d75f581546b2711d344beeaae66&run=1&cached=0) (picture elements with art direction)
- [netflix.com](https://webpagetest.org/custom_metrics.php?test=210630_BiDc7C_4aa597391e609ba73abe5d208a9853fb&run=1&cached=0) (hero does srcset with w descriptors)
- [mailchipm.com](https://webpagetest.org/custom_metrics.php?test=210630_BiDc8J_03f2588b2180940262e8f0ad38eed707&run=1&cached=0) (shows interaction with lazyloading, which... maybe I should report explicitly on that (or whether the img generated a request), too, so we don't just have to infer from zero dimensions)
- [cnn.com](https://webpagetest.org/custom_metrics.php?test=210630_BiDcE0_a51708410bc5e0cd0d005758b9a444cb&run=1&cached=0) (shows interaction with complex javascript-driven responsive loading and 1x1 trackers, which is all fine)

rviscomi:
This looks great, thanks for the test cases. I'm comfortable merging this as-is in the interest of time. If there's any follow-up feedback or bug fixes we could open another PR.

PR #221, opened 2021-06-30
Performance 2021 custom metrics

Nithanaroy:
Progress on https://github.com/HTTPArchive/almanac.httparchive.org/issues/2148

https://www.webpagetest.org/jsonResult.php?test=210630_AiDcW5_45f47b83f50f2cfd80d460827d845164&pretty=1 is the test result JSON with `lcp_stats` as the custom metric.

### DIFF
--- a/custom_metrics/almanac.js
+++ b/custom_metrics/almanac.js
@@ -710,30 +710,5 @@ return JSON.stringify({
 
   'length_of_h1s': (() => {
     return [...document.querySelectorAll('h1')].map(node => node.innerText.length);
-  })(),
-
-  'lcp_elem_stats': (() => {
-    return new Promise((resolve) => {
-      new PerformanceObserver((entryList) => {
-        const lcpCandidates = entryList.getEntries();
-        const naiveLcpEntry = lcpCandidates[lcpCandidates.length - 1];
-        resolve(naiveLcpEntry);
-      }).observe({ type: "layout-shift", buffered: true });
-    }).then(({ startTime, element, url, size, loadTime, renderTime }) => {
-      let attributes = [];
-      for (let index = 0; index < element.attributes.length; index++) {
-        const ele = element.attributes.item(index);
-        attributes[index] = { name: ele.name, value: ele.value };
-      }
-      return {
-        startTime,
-        nodeName: element.nodeName,
-        url,
-        size,
-        loadTime,
-        renderTime,
-        attributes,
-      };
-    });
-  })(),
+  })()
 });
--- a/custom_metrics/almanac.js
+++ b/custom_metrics/almanac.js
@@ -711,4 +711,29 @@ return JSON.stringify({
   'length_of_h1s': (() => {
     return [...document.querySelectorAll('h1')].map(node => node.innerText.length);
   })(),
+
+  'lcp_elem_stats': (() => {
+    return new Promise((resolve) => {
+      new PerformanceObserver((entryList) => {
+        const lcpCandidates = entryList.getEntries();
+        const naiveLcpEntry = lcpCandidates[lcpCandidates.length - 1];
+        resolve(naiveLcpEntry);
+      }).observe({ type: "layout-shift", buffered: true });
+    }).then(({ startTime, element, url, size, loadTime, renderTime }) => {
+      let attributes = [];
+      for (let index = 0; index < element.attributes.length; index++) {
+        const ele = element.attributes.item(index);
+        attributes[index] = { name: ele.name, value: ele.value };
+      }
+      return {
+        startTime,
+        nodeName: element.nodeName,
+        url,
+        size,
+        loadTime,
+        renderTime,
+        attributes,
+      };
+    });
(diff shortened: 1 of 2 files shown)


Nithanaroy:
Credits: https://gist.github.com/siakaramalegos/7cc8c0a790f8dfe0592c77cbca8440fa

custom_metrics/almanac.js
```diff
@@ -711,4 +711,29 @@ return JSON.stringify({
   'length_of_h1s': (() => {
     return [...document.querySelectorAll('h1')].map(node => node.innerText.length);
   })(),
+
+  'lcp_elem_stats': (() => {
+    return new Promise((resolve) => {
```
rviscomi, line 716: WebPageTest supports Promises in custom metrics, but I don't think it would work as implemented. The issue is that the almanac.js custom metric is an object and we're setting the value of `lcp_elem_stats` to a promise, but WPT doesn't know to wait for this specific promise to resolve.

Instead, I think we would either need to alter the response object to be a Promise that resolves when all inner Promises resolve, or (preferably) move this to its own custom metric file, for example `performance.js`.

rviscomi (changes requested):

custom_metrics/almanac.js
rviscomi: How about this for the contents of `performance.js`:

```js
//[performance]

function getLcpElement() {
  return new Promise((resolve) => {
    new PerformanceObserver((entryList) => {
      const lcpCandidates = entryList.getEntries();
      const naiveLcpEntry = lcpCandidates[lcpCandidates.length - 1];

      resolve(naiveLcpEntry);
    }).observe({ type: "largest-contentful-paint", buffered: true });
  }).then(({ startTime, element, url, size, loadTime, renderTime }) => {
    let attributes = [];
    for (let index = 0; index < element.attributes.length; index++) {
      const ele = element.attributes.item(index);
      attributes[index] = { name: ele.name, value: ele.value };
    }

    return {
      startTime,
      nodeName: element.nodeName,
      url,
      size,
      loadTime,
      renderTime,
      attributes,
    };
  });
}

return Promise.all([getLcpElement()]).then(lcp_elem_stats => {
  return {
    lcp_elem_stats
  };
});
```

Here's a test result for nytimes.com: https://webpagetest.org/custom_metrics.php?test=210630_AiDcEC_a6134c84134bd627a9522584290470ea&run=1&cached=0

```
{
  "lcp_elem_stats": [
    {
      "startTime": 2928.899,
      "nodeName": "IMG",
      "url": "https://static01.nyt.com/images/2021/06/29/nyregion/29nymayor-rcv-1/merlin_189823446_9e5b64eb-345a-4429-b618-8e9135b1c25d-threeByTwoSmallAt2X.jpg?format=pjpg&quality=75&auto=webp&disable=upscale",
      "size": 65330,
      "loadTime": 2893,
      "renderTime": 2928.899,
      "attributes": [
        {
          "name": "class",
          "value": "css-hdqqnp"
        },
        {
          "name": "src",
          "value": "https://static01.nyt.com/images/2021/06/29/nyregion/29nymayor-rcv-1/merlin_189823446_9e5b64eb-345a-4429-b618-8e9135b1c25d-threeByTwoSmallAt2X.jpg?format=pjpg&quality=75&auto=webp&disable=upscale"
        },
        {
          "name": "alt",
          "value": "New results had suggested Eric Adams's lead had narrowed. The tally was taken down hours later."
        },
        {
          "name": "role",
          "value": "presentation"
        },
        {
          "name": "loading",
          "value": "lazy"
        }
      ]
    }
  ]
}
```

And this data aligns with the LCP calculated by WPT: https://webpagetest.org/vitals.php?test=210630_AiDcEC_a6134c84134bd627a9522584290470ea&run=1&cached=0#lcp

cc @siakaramalegos

Nithanaroy:
Thanks a lot Rick and sorry for the last minute submission.

- How does WPT know to run this new file? I got some hints from the Makefile but just curios to know how this repo works.
- And where does this new `performance` key appear? `httparchive.pages`?

custom_metrics/almanac.js
siakaramalegos: Thanks so much! This worked for me on WPT as well.

rviscomi:
>How does WPT know to run this new file? I got some hints from the Makefile but just curios to know how this repo works.

WPT runs all files under the `custom_metrics` directory that end in *.js

>And where does this new performance key appear? httparchive.pages?

That's right. In the `pages` JSON blob there will be a top-level property for each custom metric.

rviscomi (approved):

Nithanaroy:
Rick, assuming this is the only metric in performance, we could have directly returned the value `getLcpElement()` right? That is `Promise.all([])` was mainly created for future proofing the schema right?

```js
// [performance]

function getLcpElement() {
  return new Promise((resolve) => {
    ...
}

return getLcpElement()

```

rviscomi:
Exactly

PR #224, opened 2021-07-13
Detect web-vitals.js

rviscomi:
Augment the `performance` custom metric with a function that detects the presence of web-vitals.js based on patterns in its response body

cc @philipwalton

### DIFF
--- a/custom_metrics/performance.js
+++ b/custom_metrics/performance.js
@@ -30,7 +30,7 @@ function getLcpElement() {
 }
 
 function getWebVitalsJS() {
-    const webVitalsJSPattern = /webVitals[\s\S]+8999999999999[\s\S]+1e12[\s\S]+(largest-contentful-paint|first-input|layout-shift)/m;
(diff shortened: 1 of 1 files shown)


custom_metrics/performance.js
```diff
@@ -27,8 +29,16 @@ function getLcpElement() {
     });
 }
 
+function getWebVitalsJS() {
+    const webVitalsJSPattern = /webVitals[\s\S]+8999999999999[\s\S]+1e12[\s\S]+(largest-contentful-paint|first-input|layout-shift)/m;
```
philipwalton, line 33: The `webVitals` part at the beginning will not necessarily be in response body, depending on which build people use (I think only the UMD and IIFE versions include it).

Also, I think having the `largest-contentful-paint|first-input|layout-shift` part after the `8999999999999[\s\S]+1e12[\s\S]` part will probably work, but I'm not 100% sure it will always be after, depending on how people are building the library (e.g. if they're building their own custom version that imports the modules directly).

Can we test for two RegExp matches?

philipwalton (changes requested):

custom_metrics/performance.js
rviscomi: Updated the regex to remove the `webVitals` pattern and to handle either ordering `(ab|ba)`.

philipwalton (approved):
👍

PR #225, opened 2021-07-16
Updated Images custom metric to use IntersectionObserver

kevinfarrugia:
Updated the Images custom metric `inViewport` property to use InteresctionObserver to improve reliability and support for:
- RTL languages
- Scrollable elements (such as carousels)

Inspired by: https://github.com/WPO-Foundation/webpagetest-docs/pull/36

Using IntersectionObserver adds some more complexity because of its async nature, however I tried to keep the custom metric as similar to its previous implementation as possible.

**Sample WPT**
https://webpagetest.org/custom_metrics.php?test=210716_AiDc99_be18dd00ed31c02ef0277471152b0fa8&run=1&cached=0

```
[
  {
    "url": "https://placekitten.com/304/304",
    "width": 304,
    "height": 304,
    "naturalWidth": 304,
    "naturalHeight": 304,
    "loading": "lazy",
    "decoding": null,
    "inViewport": true
  },
  {
    "url": "https://placekitten.com/305/305",
    "width": 305,
    "height": 305,
    "naturalWidth": 0,
    "naturalHeight": 0,
    "loading": "lazy",
    "decoding": null,
    "inViewport": false
  }
]
```

### DIFF
--- a/custom_metrics/Images.js
+++ b/custom_metrics/Images.js
@@ -1,43 +1,60 @@
-var wptImages = function (win) {
-  var images = [];
-  if (win) {
-    var doc = win.document;
-    var elements = doc.getElementsByTagName("*");
-    for (var i = 0; i < elements.length; i++) {
-      var el = elements[i];
-      if (el.tagName == "IMG") {
-        var url = el.currentSrc || el.src;
+return new Promise((resolve) => {
+  let observer = new IntersectionObserver((entries, observer) => {
+    observer.disconnect();
+    const images = entries.map((e) => {
+      const el = e.target;
+      let url = el.currentSrc || el.src;
 
-        // Only include HTTP(S) URLs i.e. skip dataURIs
-        if (url.indexOf("http") === 0) {
-          images.push({
-            url: url,
-            width: el.width,
-            height: el.height,
-            naturalWidth: el.naturalWidth,
-            naturalHeight: el.naturalHeight,
-            loading: el.getAttribute("loading"),
-            decoding: el.getAttribute("decoding"),
-            inViewport:
-              el.getBoundingClientRect().bottom >= 0 &&
-              el.getBoundingClientRect().right >= 0 &&
-              el.getBoundingClientRect().top <= window.innerHeight &&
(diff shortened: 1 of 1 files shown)


custom_metrics/Images.js
```diff
@@ -1,43 +1,60 @@
...
 
-return JSON.stringify(wptImages(window));
+    return images;
+  };
+
+  const imgs = wptImages(window);
+
+  for (let i = 0; i < imgs.length; i++) {
+    observer.observe(imgs[i]);
+  }
```
rik, line 59: If a page has no images (unlikely but the web is so vast), the script will timeout because the promise never gets resolved. (I'm also switching to for-of because it's less error prone)

```suggestion
  if (imgs.length == 0) {
    return resolve([]);
  }
  for (let img of imgs) {
    observer.observe(img);
  }
```

rik (commented):
Is there a guarantee from `IntersectionObserver` that all observed elements will be delivered in one callback? I have not studied the specification deep enough to be sure of that. If there's no guarantee, you might miss some images compared to the previous version.

kevinfarrugia:
> Is there a guarantee from `IntersectionObserver` that all observed elements will be delivered in one callback? I have not studied the specification deep enough to be sure of that. If there's no guarantee, you might miss some images compared to the previous version.

Converted to draft until I am able to clarify/confirm this.

rviscomi:
@kevinfarrugia any update on this?

kevinfarrugia:
@rviscomi I had manually tested it on a number of pages and the results are correct. I couldn't find anything detailing whether all entries should be included in the first callback or not, so I suspect it may be up to the browser.

rviscomi:
Ok thanks. Is this PR ready for review then? If so I'll try to get it reviewed and merged in time for the September crawl.

kevinfarrugia:
Yes it is ready for review but I am wary of merging it until I have a definite confirmation on the behavior of IntersectionObserver.

rviscomi:
Ok so let's leave this in draft until you confirm.

rviscomi:
Closing for now. Feel free to reopen when ready.

PR #226, opened 2021-08-02
New metric to detect service workers in pwa.js

demianrenzulli:
This PR contains the following changes to pwa.js:

- Added  a new field `serviceWorkerInfo` that uses a new regular expression to detect service worker registrations, that  matches the cases that the original one was ignoring: for example relative paths and variables used as the service worker URL. [Here is](https://webpagetest.org/result/210802_BiDcAB_61a77c283d1cdc85b764aa97658e7a2d/?f=json) an example test for mobile.twtiter.com, which was originally ignored and now it's found under `serviceWorkerInfo`.
- Added the new variable `serviceWorkerInfo` to the conditions for the `serviceWorkerHeuristic`. Ideally, the number of detected service workers should approximate more closely to `serviceWorkerHeuristic`.
- Added the `fetch` event listener to `swEventListenersInfo`. We initially removed the "fetch" keyword to avoid false positives (since the event can be used in pages too), but we recently realized that the pattern: `addEventListener("fetch", ...)`, can only happen inside a service worker, so it makes sense to add it, at least there.

We should discuss if the original `serviceWorkers` continues making sense or not. Maybe after seeing the data, we'll have a better idea.

Unfortunately, these changes won't make it for the 2021 version of Web Almanac, but we thought about leaving them prepared for 2021.

cc // @tunetheweb @rviscomi

### DIFF
--- a/custom_metrics/pwa.js
+++ b/custom_metrics/pwa.js
@@ -1,13 +1,13 @@
 //[pwa]
 const response_bodies = $WPT_BODIES;
 const requests = $WPT_REQUESTS;
-const serviceWorkerRegistrationStrictPattern = /navigator\.serviceWorker\.register\(['"]([^"']+)/m;
+const serviceWorkerStrictRegistrationPattern = /navigator\.serviceWorker\.register\(['"]([^"']+)/m;
 
 const serviceWorkerURLs = response_bodies.filter(har => {
-  return serviceWorkerRegistrationStrictPattern.test(har.response_body);
+  return serviceWorkerStrictRegistrationPattern.test(har.response_body);
 }).map(har => {
   const base = new URL(location.href).origin;
-  const serviceWorkerPath = har.response_body.match(serviceWorkerRegistrationStrictPattern)[1];
+  const serviceWorkerPath = har.response_body.match(serviceWorkerStrictRegistrationPattern)[1];
   return new URL(serviceWorkerPath, base).href;
 }).reduce((set, url) => {
   set.add(url);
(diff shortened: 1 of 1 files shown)


custom_metrics/pwa.js
```diff
@@ -124,6 +129,7 @@ return {
   swRegistrationPropertiesInfo: Object.fromEntries(swRegistrationPropertiesInfo),
   windowEventListenersInfo: Object.fromEntries(windowEventListenersInfo),
   windowPropertiesInfo: Object.fromEntries(windowPropertiesInfo),
+  serviceWorkerInfo: Object.fromEntries(serviceWorkerInfo),
   //Experimental field: Heuristic to detect if a site has a service worker even if the 'serviceWorkers' field is empty (false positives).
-  serviceWorkerHeuristic: !isObjectKeyEmpty(serviceWorkers) || !isObjectKeyEmpty(workboxInfo) || !isObjectKeyEmpty(swEventListenersInfo) || !isObjectKeyEmpty(swMethodsInfo)
+  serviceWorkerHeuristic: !isObjectKeyEmpty(serviceWorkers) || !isObjectKeyEmpty(serviceWorkerInfo) || !isObjectKeyEmpty(workboxInfo) || !isObjectKeyEmpty(swEventListenersInfo) || !isObjectKeyEmpty(swMethodsInfo)
```
tunetheweb, line 134: We discussed that having a non-empty serviceWorkers object (i.e. that a ServiceWorker was registered AND the URL that was registered was downloaded) would be a strong indicator for this heuristic.

The others were weaker indicators (e.g. if registration code is commented out or never called) and maybe two or more of them should be required to avoid false positives.

You were going to include that enhancement in this PR I thought? But doesn’t look like you did. Did you change your mind here?

custom_metrics/pwa.js
```diff
@@ -80,13 +80,18 @@ function getInfoForPattern(regexPattern, extractMatchingGroupOnly) {
   });
 }
 
+// Unlike serviceWorkerStrictRegistrationPattern that only matches SW registration scripts that contain URLs,
+// serviceWorkerLaxRegistrationPattern matches any call to the SW registration script (e.g. passing a variable, etc).
+const serviceWorkerLaxRegistrationPattern = /navigator\.serviceWorker\.register\(([^)]*)\)/g;
+const serviceWorkerInfo = getInfoForPattern(serviceWorkerLaxRegistrationPattern, true);
```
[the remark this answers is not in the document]
demianrenzulli: I changed it to `serviceWorkerRegistration`, since serviceWorkerRegistration  might be confused with the [ServiceWorkerRegistration interface](https://developer.mozilla.org/en-US/docs/Web/API/ServiceWorkerRegistration).

custom_metrics/pwa.js
demianrenzulli: I just did an update, with a function `calculateServiceWorkerHeuristic()` that returns true if either:

- `serviceWorkers` field is non empty. 
- 2 out of 4 of `serviceWorkerRegistrationInfo`, `workboxInfo`, `swEventListenersInfo`, `swMethodsInfo` are non-empty.

**Tests:**

- In [Twitter's case](https://webpagetest.org/result/210802_BiDcYP_3db14f75fc8722635c8b98cce5ee7a5c/?f=json), it returns true (Twitter has a SW).
- In [UnoTV's case](https://webpagetest.org/result/210802_AiDcNZ_e227d18859b34d1a30d05f0db5613617/?f=json), its false (UnoTV doesn't have a SW).

tunetheweb: Nice work!

tunetheweb (approved):

rviscomi (approved):

rviscomi:
I've deployed this metric to the HA server so we should have ~95% coverage in the August dataset.

PR #227, opened 2021-08-05
Add `prefers-color-scheme` custom metric

tomayac:
Detects if sites make use of `prefers-color-scheme`. 

Limitations:
- Does not detect when a site works purely with `matchMedia('(prefers-color-scheme: dark)').media` since this is (ab)used a lot for tracking without actually changing the appearance. The legitimate use case would be to set a `"dark"` class (or whatever the name may be) if and only if the media query matches, but it's hard to detect this.


tunetheweb:
Can’t you use WPT_BODIES to get around the cross-origin issue?

See https://github.com/HTTPArchive/legacy.httparchive.org/blob/master/custom_metrics/pwa.js as an example.

tomayac:
> Can’t you use WPT_BODIES to get around the cross-origin issue?
 
I can. Let me add this as a fallback.

tunetheweb:
Btw I presume you know there’s a blink feature for this?: https://chromestatus.com/metrics/feature/timeline/popularity/3868

Not sure what the difference is between that and this custom metric - and why it’s so high on that graph! Is it due to frameworks technically supporting this even if it isn’t used?

demianrenzulli:
> Not sure what the difference is between that and this custom metric - and why it’s so high on that graph! Is it due to frameworks technically supporting this even if it isn’t used?

- One of the reasons why the Chrome Status one might be showing such a high % of page loads is because it might be detecting users optin-in to Dark Mode, instead of sites implementing `prefers-color-scheme`.
- If that's not the case, and Chrome Platform Status actually detects the occurrences of the media query, another reason for the high % might be related to sites using popular 3rd party scripts that contain the `prefers-color-scheme` somewhere in the code but are not actually using the media query. 

@tomayac found an example from a site that was using this [Facebook script](https://static.xx.fbcdn.net/rsrc.php/v3ipZd4/yo/l/sv_SE/ryu30fhhn9u.js?_nc_x=Ij3Wp8lg5Kz), which contains `prefers-color-scheme` maybe for tracking purposes (but doesn't actually use it one would expect).

This custom metric would help us clear the first assumption, and focus on the second case to obtain sites actually using the media query or find suspicious behaviors leading to false positives like the one mentioned before.

tomayac:
Using `$WPT_BODIES` as a fallback now, @tunetheweb. Also asked for your review.

custom_metrics/prefers-color-scheme.js
```diff
@@ -0,0 +1,53 @@
+// [prefersColorScheme]
+
+const PREFERS_COLOR_SCHEME_REGEXP =
+  /@media\s+\(\s*prefers-color-scheme\s*:\s*(?:dark|light)\s*\)\s*\{[^\}]*\}/gms;
+
+// Checks in three passes:
+// 1. The CSS rules of same-origin and inlined stylesheets in `try {}`.
+// 2. The CSS rules of cross-origin stylesheets in `catch {}`.
+// 3. The `link[media]` attribute of conditionally loaded stylesheets in the
+//    ternary expression if steps 1. and step 2. both return `false`.
```
tomayac, line 10:
```suggestion
//    ternary expression if step 1. and step 2. both return `false`.
```

custom_metrics/prefers-color-scheme.js
```diff
@@ -0,0 +1,53 @@
+// [prefersColorScheme]
+
+const PREFERS_COLOR_SCHEME_REGEXP =
+  /@media\s+\(\s*prefers-color-scheme\s*:\s*(?:dark|light)\s*\)\s*\{[^\}]*\}/gms;
+
+// Checks in three passes:
+// 1. The CSS rules of same-origin and inlined stylesheets in `try {}`.
```
rviscomi, line 7: Could this be simplified to only use the WPT_BODIES approach? The vast majority of pages in the corpus include some kind of cross-origin stylesheet (80%) so this metric would fall back to the `catch` block most of the time. For the 20% of pages that have no cross-origin stylesheets, is there any advantage to the `document` approach? If it's all the same it'd be good to simplify the code.

```sql
SELECT
  COUNT(DISTINCT IF(type = 'css' AND NET.HOST(url) != NET.HOST(page), pageid, NULL)) / COUNT(DISTINCT pageid) AS pct
FROM
  `httparchive.summary_requests.2021_07_01_mobile`
JOIN
  (SELECT url AS page, pageid FROM `httparchive.summary_pages.2021_07_01_mobile`)
USING (pageid)
```

_(BigQuery doesn't have a `NET.ORIGIN` function so cross-host is used here as a lower bound, but I expect it'll be very close)_

rviscomi (changes requested):

custom_metrics/prefers-color-scheme.js
```diff
@@ -0,0 +1,53 @@
+// [prefersColorScheme]
```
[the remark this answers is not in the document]
tomayac: Done: https://github.com/HTTPArchive/legacy.httparchive.org/pull/227/commits/e6896e9fd39cd7035f1edc4b87c0ac014f4a3991.

custom_metrics/prefers-color-scheme.js
tomayac: I was under the impression the `document` was faster, but given the number of cross-origin stylesheets (thanks for sharing this) have now moved everything to use `$WPT_BODIES` instead (plus a second check to see if `link[rel="stylesheet"][media*="(prefers-color-scheme)"]` matches.

custom_metrics/css.js
```diff
@@ -1,30 +1,62 @@
...
+  prefersColorScheme:
+    // For all stylesheets…
+    bodies
+      .filter((request) => request.type === 'Stylesheet')
+      // …check if the `prefers-color-scheme` RegExp matches…
+      .map((request) =>
+        PREFERS_COLOR_SCHEME_REGEXP.test(request.response_body || ''),
+      )
+      // If even just one of the stylesheets matches, return `true`.
+      .filter((usesPrefersColorScheme) => usesPrefersColorScheme).length > 0 ||
```
rviscomi, line 54: Could these three methods be combined into one?
```suggestion
    // …check if the `prefers-color-scheme` RegExp matches…
    // If even just one of the stylesheets matches, return `true`.
    bodies.some((request) => {
      return request.type === 'Stylesheet' && PREFERS_COLOR_SCHEME_REGEXP.test(request.response_body || '')
    }) ||
```

custom_metrics/css.js
```diff
@@ -1,30 +1,62 @@
...
+    // If none of the stylesheets match, alternatively check if any of the
+    // stylesheet `link`s load conditionally based on `prefers-color-scheme`.
+    Array.from(
+      document.querySelectorAll(
+        'link[rel="stylesheet"][media*="prefers-color-scheme"]',
+      ),
+    ).length > 0,
```
rviscomi, line 61: NodeLists support the `length` property

```suggestion
    document.querySelectorAll('link[rel="stylesheet"][media*="prefers-color-scheme"]').length > 0,
```

rviscomi (changes requested):

rviscomi (approved):

PR #258, opened 2022-02-24
Fix error where window should be self

tomayac:


rviscomi (approved):

rviscomi:
@pmeenan what's the current process to update custom metrics on the WPT agents? I'm not sure if we should continue to use this repo and the legacy script to sync custom metrics on the test server, or if we should use the new [custom-metrics](https://github.com/HTTPArchive/custom-metrics) repo. If the latter we should also port over @tomayac's fugu-apis.js generator script.

pmeenan:
@rviscomi Right now it is still coming from the server but I'll be switching over soon (I can do it before the crawl starts if it would help).  Either way, the custom-metrics repo will need to be up to date in the next couple of weeks.

